PR #20, opened 2017-03-23
Changing app's package.json `name` to fix electron-builder issue in linux

bochaco:
The `~/.local/share/applications/maidsafe-safe-browser.desktop` contains the following `exec` command:
`Exec=/tmp/.mount_Yb2ArI/usr/bin/SAFEBrowser %u`

But the binary generated on linux is named `safe-browser`, e.g. `/tmp/.mount_Yb2ArI/usr/bin/safe-browser`

The `electron-builder` picks the app's package.json `name` rather than the `productName`.

### DIFF
--- a/app/package.json
+++ b/app/package.json
@@ -1,5 +1,5 @@
 {
-  "name": "safe-browser",
+  "name": "SAFEBrowser",
   "productName": "SAFEBrowser",
   "description": "A browser for the SAFE Network, powered by Beaker.",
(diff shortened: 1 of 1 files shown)


bochaco:
This is the incorrect solution, see this [PR](https://github.com/maidsafe/beaker-plugin-safe-authenticator/pull/61) in the authenticator plugin for the other proposed solution.

PR #81, opened 2017-07-19
fix/issue_#73: resolve issue #73

### LINKED ISSUE #73: index.html not automatically read or added
Operation System: Win7
Beaker Version: 0.3.0

When typing in a URL to a site, for example, safe://portlandspirit, the URL is rewritten so that it becomes safe://portlandspirit/index.html  This is ugly.  It should not rewrite the URL and read index.html automatically.

When clicking on a link that does not include index.html, the index.html is not added as it was when the URL is typed.  Instead, a "ERR_NOT_IMPLEMENTED" error is displayed.  index.html should not need to be in the URL, but it does need to be, then it should be automatically added when missing so it is consistent with the behavior that happens when you type an URL

shankar2105:
fix #73 - resolve index.html not automatically read.
Error page text to support HTML entities.

This fix depended on safe_app_nodejs PR [101](https://github.com/maidsafe/safe_app_nodejs/pull/101)

### DIFF
--- a/app/lib/error-page.js
+++ b/app/lib/error-page.js
@@ -1,7 +1,7 @@
 export default function (err) {
   return `<body style="font-family: sans-serif;color: #666;font-size:21px">
     <div style="margin: 35px;">
-      <h1 style="color: #555">This site can\'t be reached</h1>
+      <h1 style="color: #555">This site can&apos;t be reached</h1>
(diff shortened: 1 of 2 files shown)


Viv-Rajkumar:
CI is currently not functional in this repo, so ignoring build status from appveyor

PR #96, opened 2017-08-03
Save state to network for [MAID-2121] [WIP]

joshuef:
Biiiiig set of changes that incorporate the new MD apis from beaker-plugin-safe-app.
Refactors bookmarks/history apis as a consequence of this (removing the pointless ImmutableJS from these reducers (imo)... who added that? 😛  )

Will trigger an auth on browser load, and pull configs after that.
Saving on configs occurs on quitting the app.

data is saved to the app home container, and is a simple `JSON.stringify` of the whole state to avoid mutation complexity / excessive PUTs etc. (This could be refactored to be more fine grained if needed).

extras:
- adds a handy `logInRenderer` method to push things to the shell window
- enabled shellWindow inspection in production (hidden in a window submenu labelled 'devtools')
- opens the shellwindow inspector automatically in dev mode


joshuef:
Currently looks like the `before-quit` hook isn't being captured in linux, so looking into something more robust there.

app/background-process/safe-storage/utils/app_utils.js
```diff
@@ -0,0 +1,79 @@
...
+export const parseUrl = (url) => (
+  (url.indexOf('safe-auth://') === -1) ? url.replace('safe-auth:', 'safe-auth://') : url
+);
+
+export const deserialiseArray = (str) => {
+  let arrItems = str.split(',');
+  return Uint8Array.from(arrItems);
+}
+
+export const genKeyPair = () => {
```
bochaco: @joshuef , I think you are not using this functions (`genKeyPair`, `encrypt` and `decrypt`) so you can presumably remove them. Even if you are using them, I think you should use the ones provided by safe_app, this is something pending in the email app, to replace them and use the ones from safe_app.

josh-wilson-sociomantic: 👍  I've a branch with this removed. Got on okay this morning after we spoke about the `MD.encrypt` etc. 

I'll probably supersede this PR with another with encrypt sorted. 

Marking this as WIP for now 👍

joshuef: Actually I've a branch intending to supersede this with these removed.

I've moved over to `md.encrypt` etc as we spoke about earlier.

I'll mark this as WIP for now until it's updated / replaced.

PR #107, opened 2017-08-13
**WIP** MAID-2109 feat/logger: winston logger integration

krishnaIndia:
winston logger integrated. winston instance is passed to the plugins in the register function


bochaco:
I'm closing this as this seems to be for our beaker browser. If the same enhancement is needed for Peruse, let's adapt the code for it and send a new PR.

PR #123, opened 2017-08-23
MAID-2261 update app access api

shankar2105:
Updated UI to handle app access for Data request.
Renamed Apps root container as Apps own container.
MAID-2276 - Fixed App permission empty list on auth popup.


app/shell-window/ui/navbar.js
```diff
@@ -331,9 +331,9 @@ function showSafeAuthPopup(reqType) {
   var listCont = null;
   var noContainerDesc = 'Application is requesting for reading public unencrypted data on your behalf.';
   var ownContainerInfo = {
-    name: 'Apps root container',
+    name: 'Apps own container',
```
bochaco: We are missing an apostrophe here, I think it should read "App's own container"

PR #163, opened 2017-09-12
fix/MAID-2335: Browser shows default list of bookmarks/history before…

safesurfer:
… logging in

### DIFF
--- a/app/background-process/safe-storage/reducers/bookmarks.js
+++ b/app/background-process/safe-storage/reducers/bookmarks.js
@@ -10,18 +10,7 @@ import ACTION_TYPES from '../actions/action_types';
 import store from '../store'
 import { createActions } from 'redux-actions'
 
-const initialBookmarkState = [
-   {
-    url: 'https://safenetforum.org/',
-    title : "Safenet Forum",
-    num_visits : 0
-  },
-   {
-    url: 'safe://dir.yvette/',
-    title : "SAFE Network Directory",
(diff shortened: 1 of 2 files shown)


theWebalyst:
Sounds interesting - I think it is a pain to be force to login / create account whenever you start the browser, and hope this is the solution :)

bochaco:
@theWebalyst , I don't think this changes what you seem to be suggesting, so can you please clarify?
Currently, the browser doesn't keep the history/bookmarks locally anymore since the feature to save the state in the network was implemented.
MAID-2335 then states that the default list of bookmarks and history shown even before you log in is then incorrect as that's not what is being saved in the network.

theWebalyst:
@bochaco Ah sorry, I misunderstood what this does - I thought it was making it local so that you don't have to log in. I think forcing login is not a good UX. Although I do understand why it is better for privacy if this is not stored locally. Perhaps there could be an option to allow a local copy - but either way I think it would be good not to force login, so a user can start the browser and just use it to view websites without being logged in, as was previously the case.

bochaco:
That's fine, I imagined that's what you were thinking of. Although, note you don't need to log in to be able to browser safe sites.

theWebalyst:
Ah OK, it's just that when you start the browser it presents you with a login - it didn't occur to me ignore it! Maybe that says something about my psychology :-), but I think most people will think the same. Thanks.

Issue #162, opened 2017-09-11
safeApp.authorise() causes uncaught exception in electron


theWebalyst:
Operation System: Debian 8/Jesse 
Safe Browser Version: 0.5.1

I'm getting an intermittent problem calling safeApp.authorise() which looks like an error in the browser itself. This seems similar to another issue (which I can't find the number for right now, but as the error info is different I'm logging it separately).

The authorise() call seems to go through (because auth request is presented by the browser), but a modal dialog appears over the browser window with the following text:

```
Uncaught Exception:
TypeError: Cannot read property 'webContents' of null
	at EventEmitter.electron.ipcMain.on (/home/mrh/safe-releases/2017-09-07_Test_19_update/safe-browser-v0.5.1-linux-x64/resources/app/background-process.build.js:1821:6)
	at emitTwo (events.js:106:13)
	at EventEmitter.emit (events.js:194:7)
	at WebContents.<anonymous> (/home/mrh/safe-releases/2017-09-07_Test_19_update/safe-browser-v0.5.1-linux-x64/resources/electron.asar/browser/api/web-contents.js:247:13)
	at emitTwo (events.js:106:13)
	at WebContents.emit (events.js:194:7)
```

The code I'm using is pretty simple, but part of a larger codebase with lots of stuff going on. You can visit it on Test 19 at safe://myfd.rsports. Load the page and then click on the blue cube top right of the web page.

theWebalyst:
Possibly the same or related issues:

https://maidsafe.atlassian.net/browse/MAID-2321
https://github.com/maidsafe/beaker-plugin-safe-authenticator/issues/137

theWebalyst:
Fixed in SAFE Browser 0.5.2 😃

PR #164, opened 2017-09-12
MAID-2321 fix/The readConfig function was creating an additional an unnecessary connection

bochaco:


theWebalyst:
Is this likely to be related to the following, or not so much given that I'm getting it every time even on a clean machine and first use of the browser? Or actually, could you try it with your build, just go to safe://myfd.rsports, then click on blue cube top right - thanks! 

https://github.com/maidsafe/safe_browser/issues/162

bochaco:
This is not the only PR related to MAID-2321, there was another PR merged some days ago which also helping to solve MAID-2321. We are testing now to see if there are more enhancements to be made. We are in fact using your safe site to test.

theWebalyst:
Cool! If you get anything that runs my site please let me know as I'd love to get hold of it ASAP :-)

bochaco:
You can give it a try with the latest code from browser's master branch. I just tried it on Linux and I'm able to connect with it, it throws an error which is not related to the SAFE API or network though.

theWebalyst:
@bochachao - I had a problem with the build - are you sure master is ok? It builds but gives an error on 'yarn start'. I'm on riot.im if you have a moment.

theWebalyst:
**Update:** I get further after 'yarn package' and running the linux-distribution binary. No unusual errors on start, nothing unexpected in the linux console, but on trying to login I'm getting ```Core error: Unexpected: Could not connect to the SAFE Network``` although my IP is correct.

theWebalyst:
**Update:** I get further after 'yarn package' and running the linux-distribution binary. No unusual errors on start, nothing unexpected in the linux console, but on trying to login I'm getting ```Core error: Unexpected: Could not connect to the SAFE Network``` although my IP is correct.

PR #180, opened 2017-09-19
MAID-2022: beaker-plugin-safe-app tests

hunterlester:


app/tests/beaker-plugin-safe-app/.gitignore
```diff
@@ -0,0 +1,2 @@
+build
+yarn.lock
```
joshuef: We shouldn't be ignoring `yarn.lock` files: https://yarnpkg.com/lang/en/docs/yarn-lock/#toc-check-into-source-control

app/tests/beaker-plugin-safe-app/package.json
```diff
@@ -0,0 +1,28 @@
+{
```
joshuef: I think we should have these dev dependencies in the `app/package.json` so they're installed automatically for developers.

We can then add a `scripts` section to this and a script to run the SAFE API test server, for eg.

This'll remove the extra yarn/package files and keep everything in one place.

app/tests/beaker-plugin-safe-app/test_modules/crypto_pub_enc_key.js
```diff
@@ -0,0 +1,43 @@
...
+  });
+
+  it.skip('encrypts with sender\'s secret key and recipient\'s public key', async () => {
+    const appHandle = await testHelpers.authoriseAndConnect();
+    const encKeyPairHandle = await window.safeCrypto.generateEncKeyPair(appHandle);
+    const pubEncKeyHandle = await window.safeCryptoKeyPair.getPubEncKey(encKeyPairHandle);
+    const secEncKeyHandle = await window.safeCryptoKeyPair.getSecEncKey(encKeyPairHandle);
+    const rawSecretKey = await window.safeCryptoSecEncKey.getRaw(secEncKeyHandle);
+    // evidently, rawSecretKey is the wrong argument to pass.
+    // QUESTION: a u64 should be passed to safe_client_libs, but i thought/
```
joshuef: @bochaco @krishnaIndia any light to shed on this? I'm not sure.

app/tests/beaker-plugin-safe-app/test_modules/mutable_data.js
```diff
@@ -0,0 +1,259 @@
...
+
+    /** manually creating Mutable Data without quickSetup, to set signing key as null *****/
+    const mdHandle = await window.safeMutableData.newRandomPublic(appHandle, testHelpers.TAG_TYPE_DNS);
+    let permissionsSetHandle = await window.safeMutableData.newPermissionSet(appHandle);
+    await window.safeMutableDataPermissionsSet.setAllow(permissionsSetHandle, 'ManagePermissions');
+    const permissionsHandle = await window.safeMutableData.newPermissions(appHandle);
+    await window.safeMutableDataPermissions.insertPermissionsSet(permissionsHandle, null, permissionsSetHandle);
+    const entriesHandle = await window.safeMutableData.newEntries(appHandle);
+    await window.safeMutableData.put(mdHandle, permissionsHandle, entriesHandle);
+    /***************************************************************************************/
```
joshuef: lets remove this comment line

app/tests/beaker-plugin-safe-app/test_modules/helpers.js
```diff
@@ -0,0 +1,115 @@
...
+  const appHandle = await window.safeApp.initialise(appInfo2);
+  const authUri = await authoriseApp(appHandle);
+  await window.safeApp.connectAuthorised(appHandle, authUri);
+  const mdHandle = await window.safeMutableData.newRandomPublic(appHandle, TAG_TYPE_DNS);
+  return window.safeMutableData.quickSetup(mdHandle, entriesObject || {});
+}
+
+const createRandomXorName = () => crypto.randomBytes(32);
+
+const createUnownedMutableData = async () => {
```
joshuef: It might be simpler/cleaner (one less thing to miss), to just do: `export const createUnownedMutableData = ` and then you don't need to reexport at the bottom of the file.

app/tests/beaker-plugin-safe-app/test_modules/mutable_data_permissions_set.js
```diff
@@ -0,0 +1,34 @@
+let should = require('should');
+let testHelpers = require('./helpers');
+
+describe('window.safeMutableDataPermissionsSet', () => {
+  // QUESTION: Please review the following statement. Is it accurate?
+  it('sets allowable actions on a Mutable Data structure by an app, represented by the app\'s signing key', async () => {
```
joshuef: I'm not sure about the `apps signing key` as being representative... @krishnaIndia @bochaco does it make sense to yous?

app/tests/beaker-plugin-safe-app/test_modules/safeApp/refreshContainersPermissions.spec.js
```diff
@@ -0,0 +1,21 @@
+let should = require('should');
+let testHelpers = require('../helpers');
+
+describe('window.safeApp.refreshContainersPermissions', () => {
+  // QUESTION: What is the purpose of refreshContainersPermissions?
```
joshuef: I guess you'd need to change them, and then pull again to get them updated? 

I'd assume the use case of an app where you request perms and then need to check if you've been given them?

@krishnaIndia can you maybe clarify here, please.

app/tests/beaker-plugin-safe-app/test_modules/mutable_data_permissions_set.js
bochaco: Yes, and the test is just missing the step to effectively insert that `PermissionsSet` into the MDs mapped to the app's sign key.

app/tests/beaker-plugin-safe-app/test_modules/safeApp/refreshContainersPermissions.spec.js
bochaco: If the permissions for an app are changed either by itself (e.g. requesting additional containers permissions), or by another app, or by another running instance of the same app, these changes are not pushed to the app, so this should allow the app to refresh them as they are kept in the safe_client_libs cache, so the cache is updated with up to date permissions.

app/tests/beaker-plugin-safe-app/test_modules/crypto_pub_enc_key.js
bochaco: There seems to be an issue there, even that the safe_app_nodejs (and its counterpart in the safe_app plugin) `PubEncKey.encrypt` documentation states that it expects either a buffer or string type for the secret key input param, it seems to be accepting only a `SecEncKey` object, therefore the test should provide the `secEncKeyHandle`, but first the safe_app plugin needs to be fixed as well as the documentation.

app/tests/beaker-plugin-safe-app/test_modules/helpers.js
hunterlester: right, I just need to figure out the proper babelify configuration in `gulpfile` to transform `import/ export`, otherwise, as it is, `import/export` throw errors since they are not recognized.

joshuef:
One tiny thing: `warning package.json: "dependencies" has dependency "express" with range "^4.15.3" that collides with a dependency in "devDependencies" of the same name with version "^4.15.4"`

And then I'm still getting errors with the updated beaker-safe-app, as noted: https://github.com/maidsafe/beaker-plugin-safe-app/pull/90

PR #193, opened 2017-09-28
fix/Extend CSP rule to allow favicon from safe and localhost protocols

bochaco:


hunterlester:
Tested on Windows with success.

PR #201, opened 2017-10-05
fix/Always take the last argument as the URI to open

bochaco:
Also, leave the protocols handlers to be registered by the plugins themselves rather than by the browser.


app/background-process/safe-storage/safenet_comm.js
```diff
@@ -8,13 +8,18 @@ import { getAPI } from './helpers';
...
+    vendor: 'MaidSafe.net Ltd'
+  },
+  initOpts: {
+    registerScheme: false
+  },
+  authOpts: {
+    own_container: true
+  },
+  permissions: {}
 };
```
joshuef: I think we need the OSX 'devmode' check here also

bochaco: In this case I don't think so, my understanding is that this is using the IPC to communicate with the authenticator, and in fact we are not even registering the scheme for this app

joshuef: Ach, aye; you're right. We're not registering it here at all, eh.

Though maybe we should/could be (although toootally out of scope of this). One of the reasons the browser doesn't use `safe-node-app` itself was the problem of getting the response from the thing sending the response. But I wonder if our new setup would make that work.... Then we could get rid of a lot of hacky imports from beaker-plugin-safe-app

PR #208, opened 2017-10-14
Safe API test added: md entries insert

loziniak:
test for a problem described in this forum topic:
https://forum.safedev.org/t/problem-with-inserting-data-to-apps-own-container/1022


hunterlester:
Here's a test, if I've read the forum post correctly, for what you are intending: https://github.com/maidsafe/safe_browser/blob/master/app/tests/beaker-plugin-safe-app/test_modules/mutable_data_mutation.spec.js#L5

loziniak:
Both approaches outlined by @bochaco seem to be already covered by unit tests. My test is a result of being misled by API documentation. I thought that calling `insert` on entries commits data to network.

PR #212, opened 2017-10-17
feat/tests: Add spectron for automated testing MAID-2403

joshuef:


bochaco:
This is for our beaker browser, so I'm closing it as this may be obsolete since we are moving onto Peruse.

PR #228, opened 2017-11-14
MAID-2418: test upgrades

hunterlester:


app/tests/beaker-plugin-safe-app/test_modules/safeApp/clearObjectCache.spec.js
```diff
@@ -0,0 +1,14 @@
+let should = require('should');
+let testHelpers = require('../helpers');
+
+describe('window.safeApp.clearObjectCache', () => {
+  it('resets the object cache kept by the underlyging library', async () => { 
+    const appHandle = await testHelpers.authoriseAndConnect();
+    
+    should(await window.safeApp.clearObjectCache(appHandle)).be.fulfilled();
```
bochaco: I would have a test which creates some other object with the app instance, let's say a `SignKeyPair`, then after you call the `clearObjectCache` check if you can still use the app's handle to do something, but you cannot use the `SignKeyPair` handle since it was invalidated and removed so you should get an error saying `invalid handle`.

PR #239, opened 2017-12-04
Resolved address bar issue

shankar2105:
Resolved address bar issue on clicking enter

### DIFF
--- a/app/shell-window/ui/navbar.js
+++ b/app/shell-window/ui/navbar.js
@@ -953,7 +953,7 @@ function onKeydownLocation (e) {
       }
 
       var selectionUrl = selection.url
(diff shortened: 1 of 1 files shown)


app/shell-window/ui/navbar.js
```diff
@@ -953,7 +953,7 @@ function onKeydownLocation (e) {
       }
 
       var selectionUrl = selection.url
-      if ((selectionUrl.subStr(-1) !== '/') && !path.extname(selectionUrl)) {
+      if ((selectionUrl.substr(-1) !== '/') && !path.extname(selectionUrl)) {
```
krishnaIndia: use `endsWith` instead of `substr(-1)`

PR #242, opened 2017-12-13
docs/Change prerequisite for node to v8.0.0

bochaco:


README.md
```diff
@@ -8,7 +8,7 @@ A browser designed to open `safe://` websites on the SAFE Network.
 
 1. Prerequisites:
 
-    * Node.js 7.10.0 (we recommend installing it via [nvm](https://github.com/creationix/nvm))
+    * Node.js 8.0.0 (we recommend installing it via [nvm](https://github.com/creationix/nvm))
```
hunterlester: I'm not sure if it will make much difference but should we set this to 8.9.x?

bochaco: I think as far as the doc is concerned we can just say >=8.0.0, then we'll decide which version to use in our CI.

PR #244, opened 2017-12-15
chore/pull crust.config from release_config

hunterlester:


hunterlester:
You'll see that the builds are failing.
They are failing because of `bad credentials` when `curl`ing from `maidsafe/release_config`, which is not an issue when I push a tag.

It may have do to with the following:
> However, secure variables are not decoded during Pull Request builds which prevents someone from submitting PR with malicious build script displaying those variables.
https://www.appveyor.com/docs/build-configuration/#secure-variables

PR #256, opened 2018-03-09
chore/safe-app plugin: Version change to v0.10.1 to upgrade safe-app plugin to v0.5.2

bochaco:


hunterlester:
Confirmed.

image

PR #262, opened 2018-06-11
[WIP] MAID-2688 chore/connectivity: enable reconnect for Peruse app and update state on network status

hunterlester:
~Why WIP? Because this task is blocking while figure out how to test the implementation without running into https://maidsafe.atlassian.net/browse/MAID-2639.~

### DIFF
--- a/app/extensions/safe/network/authenticator-comms.js
+++ b/app/extensions/safe/network/authenticator-comms.js
@@ -8,8 +8,10 @@ import {
     isCI,
     isRunningSpectronTestProcessingPackagedApp
 } from 'appConstants';
+import { SAFE } from '../constants';
 import { parse as parseURL } from 'url';
 import { addNotification, clearNotification } from 'actions/notification_actions';
+import { setNetworkStatus } from '../actions/peruse_actions';
 import { setIPCStore } from '../ffi/ipc';
 
 const queue = [];
@@ -23,6 +25,20 @@ export const authFromQueue = async () =>
     {
         authFromInternalResponse( queue[0] ); // hack for testing
     }
+};
+
+const tryConnect = async (res) => {
+  try
+  {
+    peruseAppObj = await peruseAppObj.auth.loginFromUri( res );
+    store.dispatch(clearNotification()); 
+  }
+  catch ( err )
+  {
+    setTimeout(() => {
+      tryConnect( res );
+    }, 5000);
+  }
 };
 
 export const authFromInternalResponse = async ( res, isAuthenticated ) =>
@@ -42,13 +58,14 @@ export const authFromInternalResponse = async ( res, isAuthenticated ) =>
 
             if ( err.message.startsWith( 'Unexpected (probably a logic' ) )
             {
-                message = 'Check your current IP address matches your registered address at invite.maidsafe.net';
+                message = 'Attempting to connect. Check your network connection, then verify that your current IP address matches your registered address at invite.maidsafe.net';
             }
 
             // TODO: Remove check when network is opened up
             if ( isRunningSpectronTestProcessingPackagedApp || isCI ) return;
 
             store.dispatch( addNotification( { text: message, onDismiss: clearNotification } ) );
+            tryConnect( res );
         }
 
         logger.error( err.message || err );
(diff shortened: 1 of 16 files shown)


app/extensions/safe/network/authenticator-comms.js
```diff
@@ -96,7 +97,6 @@ export const initAnon = async ( passedStore ) =>
 
     try
     {
-        // does it matter if we override?
         const networkStateCb = (state) => {
```
joshuef: @hunterlester you mean this line is not being triggered, right?

hunterlester:
Requires local link of https://github.com/maidsafe/safe_app_nodejs/pull/251 to fire network callback when unregistered connection event occurs.
Network callback fires upon connection and state dispatch is happening but I'm unable to observe updated state: https://github.com/maidsafe/safe_browser/pull/262/files#diff-c1373fedfea725334ffc2413bc67f357R88

Also for test, error is thrown when attempting to import function, https://github.com/maidsafe/safe_browser/pull/262/files#diff-a003f29f7e2f9aeecfe6e3fbb39e3d2eR8, for same reason as before where other operations in the module are initialised, one or two of which rely on a main process.

app/extensions/safe/network/authenticator-comms.js
```diff
@@ -76,6 +79,42 @@ export const handleSafeAuthAuthentication = ( uriOrReqObject, type ) =>
...
 
+export const onNetworkStateChange = (store) => (state) =>
+{
+  logger.info('net callback state: ', state);
+  const previousState = store.getState().peruseApp.networkStatus;
+  logger.info('previousState: ', previousState);
+  store.dispatch( setNetworkStatus( state ) );
+  // TODO/BOOKMARK: Expected is an updated store state but not happening
+  logger.info('store state: ', store.getState());
+  const newState = store.getState().peruseApp.networkStatus;
```
joshuef: @hunterlester the `dispatch` calls happen async, so here, when you're pinging for `getState` again, it hasn't necessarily updated the store.

But you don't need to ping the store for this anyway, the passed `state` to this function is your newState.

app/extensions/safe/network/authenticator-comms.js
```diff
@@ -76,6 +79,42 @@ export const handleSafeAuthAuthentication = ( uriOrReqObject, type ) =>
...
+      store.dispatch( addNotification(
+        {
+          text: `Network state: ${newState}. Reconnecting...`,
+          type: 'error',
+          onDismiss: clearNotification
+        } 
+      ));
+    }
+  }
+  while (isDisconnected) {
```
joshuef: I dont think this while is what's needed here (`isDisconnected` wont change in this function call...)

if you set up a func that will check the state after the timeout, and recall itself if needed, that should do the trick:
```js

  const attemptReconnect = ( store ) =>
  {
      // logger.info(' in while disconnected')
      setTimeout(() => {
          logger.info('Attempting reconnect...');
          peruseAppObj.reconnect();

          if( store.getState.peruseApp.networkStatus === SAFE.NETWORK_STATE.DISCONNECTED )
          {
              attemptReconnect(store );
          }
      }, 5000);

  }
```

hunterlester:
@joshuef Reconnect notification and operation running nicely upon `Disconnect` event.
If test is attempted, it must be linked with https://github.com/maidsafe/safe_app_nodejs/pull/252 in order to prevent false `Connected` when `app_reconnect` is called.

Almost there with the state update test. Need a little help with mocking store, see comment: https://github.com/maidsafe/safe_browser/pull/262/files#diff-a003f29f7e2f9aeecfe6e3fbb39e3d2eR84
I attempted to utilise Redux's `createStore` function but it doesn't quite replicate the same behavior as in the app's `configureStore` which relies on main process.
Reading [Jest Mock Functions](https://facebook.github.io/jest/docs/en/mock-function-api.html), the docs are unclear to me.

joshuef:
@hunterlester my missing comment was that I was still seeing the error post-reconnect when you'd been trying to load a page.

joshuef:
@hunterlester CI tests are failing here.

hunterlester:
Closing PR to iterate over CI tests on my own fork.

PR #263, opened 2018-06-13
MAID-2687 chore/auth: pass auth info object notification

hunterlester:
I think the scope of the task was covered.
I didn't get too far into polishing notification UI because I need input on basic design and layout.
Discovered that am unable to authorise `shareMDataReqUri`, see note at `TODO/BOOKMARK`.

### DIFF
--- a/app/components/Notifier/Notifier.jsx
+++ b/app/components/Notifier/Notifier.jsx
@@ -24,6 +24,7 @@ export default class Notifier extends Component
        onDismiss   : PropTypes.string,
        onAccept    : PropTypes.string,
        onDeny      : PropTypes.string,
+       reactNode    : PropTypes.object,
        updateNotification : PropTypes.func
    }
     static defaultProps =
@@ -51,11 +52,12 @@ export default class Notifier extends Component
             id,
             isPrompt,
             text,
+            reactNode,
             type,
             updateNotification
         } = this.props;
 
-        if ( !text )
+        if ( !text && !reactNode )
         {
             return ( <div /> );
         }
@@ -73,26 +75,74 @@ export default class Notifier extends Component
                 updateNotification({ id, response: 'deny' })
             };
         }
+		 
+        const reactNodeToElement = ( nodeObject ) =>
+        {
+            const nodeDescription = {};
+            Object.keys( nodeObject ).forEach( key => {
+                if( key === 'type' )
+                {
+                    return nodeDescription[key] = nodeObject[key];
+                }
 
+                if( key === 'props' )
+                {
+                    Object.keys( nodeObject[key] ).forEach( prop => {
 
+                        if( prop === 'children' )
+                        {
+                            return nodeDescription.children = nodeObject.props.children;
+                        }
+
+                        nodeDescription.props = {
+                            [prop] : nodeObject[key][prop]
+                        }
+                     } )
+
+                }
+            } );
+            if( Array.isArray( nodeDescription.children ) )
+            {
+                nodeDescription.children = nodeDescription.children.reduce((acc, val) => acc.concat(val), []).map( child => reactNodeToElement(child) );
+            } else if ( nodeDescription.children instanceof Object && !Array.isArray( nodeDescription.children ) ) {
+                nodeDescription.children = reactNodeToElement(nodeDescription.children);
+            }
+
+            const elementType = nodeDescription.type;
+            const elementProps = nodeDescription.props || null;
+            const elementChildren = nodeDescription.children || null;
+
+            return React.createElement( elementType, elementProps, elementChildren );
+        };
+
+        const reactElement = reactNode ? reactNodeToElement(reactNode) : null;
+        
         return (
             <Row hasMinHeight className={ styles.container }>
                 <MessageBox messageType={type}>
                     <Row verticalAlign="middle" align="center">
-                        <Text className={CLASSES.NOTIFIER_TEXT}>{ text }</Text>
+                        <Column key="notifier-text" align="left">
+                          {
+                            reactNode && reactElement
+                          }
+                          {
+                            !reactNode &&
+                            <Text className={CLASSES.NOTIFIER_TEXT}>{ text } </Text>
+                          }
+                        </Column>
                         {
                             handleOnAccept &&
-                            <Column align="left">
+                            <Column key="notifier-accept" align="left">
                                 <Button role="promoted" onClick={ handleOnAccept }>{ acceptText }</Button>
                             </Column>
                         }
                         {
                             handleOnDeny &&
-                            <Column align="left">
+                            <Column key="notifier-deny" align="left">
                                 <Button onClick={ handleOnDeny }>{ denyText }</Button>
                             </Column>
                         }
-                        <Column align="left">
+                        <Column key="notifier-dismiss" align="left">
                             <IconButton
                                 role="subtle"
                                 iconType="close"
--- a/app/extensions/safe/components/authRequest.js
+++ b/app/extensions/safe/components/authRequest.js
@@ -1,0 +1,90 @@
+import React, { Component } from 'react';
+import logger from 'logger';
+// TODO: When I import the following, an error is thrown in CodeMirror, used by nessie-ui, that navigator is not defined
+//import { Text, Column } from 'nessie-ui';
+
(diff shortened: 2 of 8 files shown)


app/components/AuthRequest/AuthRequest.jsx
```diff
@@ -0,0 +1,60 @@
...
+    static propTypes =
+   {
+       authReqData : PropTypes.object
+   }
+
+    render()
+    {
+        const { authReqData } = this.props;
+        logger.info('INSDIDE OF AUTHREQUERST>JS');
+        logger.info(authReqData);
```
joshuef: lets clean these up

app/components/AuthRequest/AuthRequest.jsx
```diff
@@ -0,0 +1,60 @@
...
+        if (authReqData.mDataReq) {
+          reqType = 'mDataReq'; 
+          reqTypeText = ' Requests mData Access';
+        }
+        
+        return (
+          <div>
+              <Text className='blockText'>{ authReqData[reqType].app.name + reqTypeText }</Text>
+              <Text className='blockText'>{ authReqData[reqType].app.vendor }</Text>
+              <Text className='blockText'>Requested containers:</Text>
```
joshuef: lets only show this if containers are requested

app/components/AuthRequest/AuthRequest.jsx
```diff
@@ -0,0 +1,60 @@
+// @flow
```
joshuef: This, as part of SAFE, needs to originate from the extensions. There's a couple of example APIs if we need to create a new component.

The simpler thing though, may be to just make the Notifier more generically expandable, rather than creating a new component...

app/components/Notifier/Notifier.jsx
```diff
@@ -6,6 +6,7 @@ import PropTypes from 'prop-types';
 import styles from './notifier.css';
 import logger from 'logger';
 import { CLASSES } from 'appConstants';
+import AuthRequest from '../AuthRequest';
```
joshuef: We shouldn't be importing something so specifically SAFE in the browser domain. This should be abstracted/enabled via extensions if we need this.

app/extensions/safe/ffi/ipc.js
```diff
@@ -182,6 +189,7 @@ class ReqQueue
         {
             // FIXME: if error occurs for unregistered client process next
             self.req.error = err.message;
+            // TODO/BOOKMARK: leaving off here. share MData req URI is causing error when used to call auth_decode_ipc_msg in authenticator.js
```
joshuef: Yeh we should open up a separate bug for this.

app/components/Notifier/Notifier.jsx
```diff
@@ -73,13 +75,24 @@ export default class Notifier extends Component
...
 
 
         return (
             <Row hasMinHeight className={ styles.container }>
                 <MessageBox messageType={type}>
                     <Row verticalAlign="middle" align="center">
-                        <Text className={CLASSES.NOTIFIER_TEXT}>{ text }</Text>
+                        <Column align="left">
+                          {
+                            authReqEl && authReqEl
```
hunterlester: This line throws an error that component `Column` is expecting `ReactNode` but instead was passed `children`, however, `authReqEl` is an React element object:
```
{ _owner: null,
  key: null,
  props: { children: [ [Object], [Object], [Object], [Object], [Object] ] },
  ref: null,
  type: 'div' }
```

joshuef: Yo, so here it's needed to rebuild the react components with react.createElement. There's some looping over this that needs done eg (a rough version):

```jsx
 const reReact = ( thing ) =>
        {
            const reacter = {
            };
            Object.keys( thing ).forEach( key => {
                if( key === 'type' )
                {
                    return reacter[key] = thing[key];
                }

                if( key === 'props' )
                {
                    Object.keys( thing[key] ).forEach( prop => {

                        if( prop === 'children' )
                        {
                            return reacter.children = thing.props.children;
                        }

                        reacter.props = {
                            [prop] : thing[key][prop]
                        }
                     } )

                }
            } )
            const newChildArray = [];
            let el;
            if( Array.isArray( reacter.children ) )
            {
                reacter.children.forEach( child => newChildArray.push( reReact(child) ))
                reacter.children = newChildArray;
            }

            console.log( Array.isArray( reacter.children ) , 'arrrrrrrr??????????')
            return React.createElement( reacter.type, reacter.props, reacter.children);
        }
        
        const restruct = reReact(authReqEl);
        
        console.log('and so we have:', restruct )
```

the other thing here is that instead of adding `authReq` stuff to `Notifier`, as that's SAFE specific. This should come from / be detected from the props. eg: Either there's a text prop, OR, if a 'textNode' is passed, it parses the react node if needed. (basically as you have, but a more generic naming)

app/extensions/safe/manageAuthNotifications.js
```diff
@@ -15,7 +18,61 @@ export const REQ_TYPES = {
     MDATA     : 'MDATA'
 };
 
-
+export const createAuthRequestElement = (authReqData) =>
+{
+  const R = React.createElement;
```
joshuef: @hunterlester i think it's cleaner if we stay using jsx here instead of creating the objects in this fashion. Then it's consistent across the project.

We could also then separate out the component from this file.

app/extensions/safe/manageAuthNotifications.js
```diff
@@ -15,7 +18,61 @@ export const REQ_TYPES = {
...
+    reqTypeText = ' requests mData Access';
+  }
+
+  let children =
+  [
+    R('p', { className: 'blockText', key: authReqData[reqType].app.name }, `${authReqData[reqType].app.name} ${reqTypeText}`),
+    R('p', { className: 'blockText', key: authReqData[reqType].app.vendor }, `${authReqData[reqType].app.vendor}`),
+  ];
+
+  if ( authReqData[reqType].containers )
```
joshuef: I think we need to check for length here as this is still showing even if no container do afterwards (that or the containers are missing...)

app/extensions/safe/manageAuthNotifications.js
```diff
@@ -63,18 +120,15 @@ export const addAuthNotification = ( authReqData, app, sendAuthDecision, store )
...
-    // }
+    const textNode = createAuthRequestElement(authReqData);
     const notificationId = Math.random().toString( 36 );
     const theNotification = {
         // TODO: where should ID actually be applied?
         id: notificationId,
         text,
-        isPrompt: true
+        isPrompt: true,
+        textNode
```
joshuef: I think `textNode` is maybe too specific... It could be a non text node, so maybe we just make it 'node' ? or 'reactNode' ?

app/components/Notifier/Notifier.jsx
```diff
@@ -73,26 +75,70 @@ export default class Notifier extends Component
                 updateNotification({ id, response: 'deny' })
             };
         }
+		 
+        const textNodeToElement = ( thing ) =>
```
joshuef: Let's rename `thing` to be more relevant. `nodeObject` maybe?

app/extensions/safe/components/authRequest.js
```diff
@@ -0,0 +1,116 @@
...
+
+  if (authReqData.mDataReq)
+  {
+    reqType = 'mDataReq'; 
+    reqTypeText = ' requests mData Access';
+  }
+
+  let children =
+  [
+    R('p', { className: 'blockText', key: authReqData[reqType].app.name }, `${authReqData[reqType].app.name} ${reqTypeText}`),
```
joshuef: @hunterlester let's use JSX here instead of react.createClass everywhere so it's consistent.

app/extensions/safe/components/authRequest.js
```diff
@@ -0,0 +1,116 @@
...
+      Object.keys(mdatum.perms).filter((perm) => mdatum.perms[perm]).forEach((perm, i) => {
+        children.push(R('p', { key: metaName + perm + i }, perm));
+      });
+    }); 
+  }
+
+  return R('div', {}, children);
+};
+
+//export class AuthRequest extends Component
```
joshuef: lets get rid of this commented code.

PR #264, opened 2018-06-15
MAID-2801 fix/auth: unlock icon prevents notification of previsouly authorised application

hunterlester:
Why WIP?
See my comment on line 98.

### DIFF
--- a/app/extensions/safe/ffi/authenticator.js
+++ b/app/extensions/safe/ffi/authenticator.js
@@ -25,7 +25,6 @@ import { SAFE } from 'extensions/safe/constants';
 // private variables
 const _registeredClientHandle = Symbol( 'registeredClientHandle' );
 const _nwState = Symbol( 'nwState' );
-const _reAuthoriseState = Symbol( 'reAuthoriseState' );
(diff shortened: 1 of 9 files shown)


app/extensions/safe/ffi/authenticator.js
```diff
@@ -86,6 +86,18 @@ class Authenticator extends SafeLib
...
+    get reAuthoriseState()
+    {
+      return this[_reAuthoriseState];
+    }
+
+    set reAuthoriseState( state )
+    {
+      logger.info('-----------------------------------------------');
+      logger.info('old reAuthoriseState: ', this.reAuthoriseState);
+      logger.info('new reAuthoriseState: ', state);
```
hunterlester: I'm able to observe a successful setting of `reAuthoriseState` for the `Authenticator` instance when clicking on lock icon. Subsequent clicks will correctly show the previous state.
However, icon when is unlocked, `reAuthoriseState`  has value of 1, and I authorise an app, the notification still appears because `this[_reAuthoriseState]` has an unexpected value of 0.

PR #266, opened 2018-06-28
MAID-2813 fix/electron: close hanging process upon closing window

hunterlester:


app/openWindow.js
```diff
@@ -104,6 +104,10 @@ const openWindow = ( store ) =>
         {
             browserWindowArray.splice( index, 1 );
         }
+        if ( process.platform !== 'darwin' )
+        {
+            app.quit();
```
joshuef: We need to check the window count before doing quit. 

Otherwise this is working 👍

PR #268, opened 2018-07-04
MAID-2849 fix/auth-app: invite link open single tab, builds auth-web-app and upgrades to react-router v4

hunterlester:


app/extensions/safe/webviewPreload.js
```diff
@@ -185,11 +185,6 @@ export const setupPreloadedSafeAuthApis = ( store ) =>
 
                 callbackArgs = [theCall.response];
 
-                if ( theCall.isListener )
```
joshuef: we dont have any listeners anymore?

app/reducers/tabs.js
```diff
@@ -73,14 +75,6 @@ const addTab = ( state, tab ) =>
 
     let newState = [...state];
 
-    // Prevent http tabs at all
-    // TODO. This via middleware
-    if ( tab.url.startsWith( 'http' ) )
```
joshuef: how is this happening now?

package.json
```diff
@@ -28,7 +28,7 @@
     "build-dll": "cross-env NODE_ENV=development webpack -r babel-register --config webpack.config.renderer.dev.dll.js --progress --profile --colors",
     "build-dll-live": "cross-env NODE_ENV=production webpack -r babel-register --config webpack.config.renderer.live.dev.dll.js --progress --profile --colors",
     "build-main": "cross-env NODE_ENV=production webpack -r babel-register --config webpack.config.main.prod.js --progress --profile --colors",
-    "build-auth": "cross-env NODE_ENV=production webpack -r babel-register --config app/extensions/safe/auth-web-app/webpack.config.auth-app.js --progress --profile --colors",
+    "build-auth-app": "cross-env NODE_ENV=development webpack -r babel-register --config app/extensions/safe/auth-web-app/webpack.config.js --progress --profile --colors",
```
joshuef: lets build for `production`

joshuef: We also need to add this build to the `build` command

joshuef: and automate the `yarn` for this on install too.

joshuef: @hunterlester we need this basically in a state where a fresh install will have the full auth app as needed 👍

app/extensions/safe/webviewPreload.js
hunterlester: @joshuef right, so prior I was always receiving `null` for `appList` and `auth-web-app` would crash if an app were either revoked from the `appList` for authorised to be added to the `appList`. It was just good timing before that `auth-web-app` was still functioning because `getAuthorisedApps` would be called after the `appList` listener updated the respective state with `null`.

app/reducers/tabs.js
hunterlester: @joshuef This piece is already currently handled by [blockNonSafeRequests](https://github.com/maidsafe/safe_browser/blob/peruse/app/extensions/safe/blockNonSafeReqs.js#L27). Removing this allows just one external window to be opened instead of 3 to 4 for per click.

package.json
hunterlester: @joshuef I just tried with success to run `yarn` with this PR branch on a Linux VM to successfully build `auth-web-app`. And OSX was good for you, yeah?

app/reducers/tabs.js
joshuef: 👍

app/extensions/safe/webviewPreload.js
joshuef: fair enough 👍

Issue #275, opened 2018-07-19
tab spamming on openExternal link


hunterlester:
SAFE Browser Version: head of `peruse` branch
Description:
Steps to reproduce: Click on `Claim Invitation` button in `Create Account` process

PR #276, opened 2018-07-19
Issue #275 fix/webview: properly handles http links when triggering did-fail-load events

hunterlester:


hunterlester:
@joshuef Did the whole browser crash or just the tab?
Was it just that the tab closed? That's expected. I just pushed a commit that additionally adds an error notification to that behavior.
Otherwise, I wasn't able to reproduce a browser crash on either Windows or OSX, if that was the case.

joshuef:
With this latest version @hunterlester the browser is no longer crashing, but I think closing the tab + error message is wrong here.

I think we should just be showing the error inside the tab page , similarly to 'page doesnt exist'. Let's have a look at this when you're on in a bit.

hunterlester:
@joshuef See what you think of this solution. I tried something like loading a `404.html` but that required a `loadURL` action but what we are wanting is to show the user the that page has failed to load but also keeping the tab url visible.
So here this latest commit simply manipulates the DOM upon failure of a non-existent localhost resource to load.

Issue #277, opened 2018-07-19
Clean up auth notification in `peruse` branch


hunterlester:

hunterlester:
Addressed by PR https://github.com/maidsafe/safe_browser/pull/278

PR #278, opened 2018-07-19
Issue #277 chore/styles: cleans up auth request notificaiton to only show details on hover

hunterlester:
@joshuef At first I tried placing the styles in `notifier.css`, where I noticed them not loaded when `Notifier` is rendered. There are absolutely no styles that will take effect from `notifier.css`. What is that about?

I then tried placing styles in `authReqest.css` and importing it in `authRequest.js`, for which I get parsing error related to babel, which I don't understand yet because the config in `webpack.config.base.js` looks just fine.

### DIFF
--- a/app/app.global.css
+++ b/app/app.global.css
@@ -27,3 +27,14 @@ html
     font-size: 62.5%;
     line-height: 1.2em;
 }
+
+.info_box_details
+{
+    display: none;
+    margin-top: 5%;
+}
+
+.info_box_expander:hover + .info_box_details
+{
+    display: block;
+}
--- a/app/components/Notifier/Notifier.jsx
+++ b/app/components/Notifier/Notifier.jsx
@@ -120,7 +120,7 @@ export default class Notifier extends Component
         return (
             <Row hasMinHeight className={ styles.container }>
                 <MessageBox messageType={type}>
-                    <Row verticalAlign="middle" align="center">
+                    <Row verticalAlign="top" align="center">
                         <Column key="notifier-text" align="left">
                           {
                             reactNode && reactElement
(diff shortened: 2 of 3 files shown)


hunterlester:
@joshuef The only way I can get the styles to take effect are to place them in `app.global.css`.

app/components/Notifier/notifier.css
```diff
@@ -8,3 +8,13 @@
 
     font-size: 1.4rem;
 }
+
+.hide
```
joshuef: If we're going to have a general class for hiding content, I think `hidden` is a better name.

But if we're doing _that _ it's not so normal for it to be shown (yet still have this class... [via some css trickery]).

So I think we should rename these classes. We can be more specific (even if it needs to be in app.global for now).

maybe: `more_details_button` and `more_details_details`? But that could be too specific... `info_box_details` and `info_box_expander` might work and be clear enough... and then `info_box` is something with these features...

app/app.global.css
```diff
@@ -27,3 +27,13 @@ html
     font-size: 62.5%;
     line-height: 1.2em;
 }
+
+.info_box_details
+{
+    display: none;
+}
+
+.info_box_exander:hover .info_box_details
```
joshuef: `exander` -> `expander`

joshuef:
@hunterlester this is working nicely.

But let's get more space between the details 'link' and the popup: 
https://www.dropbox.com/s/jckohhkogn936r3/Screenshot%202018-07-24%2008.41.31.jpg?dl=0

Also, can make sure that the details link is confined to the text only (ie. that you cant open up details by moving the mouse over the area to the right of the text).

Issue #279, opened 2018-07-20
Ignoring auth requests by clicking `X` prevents further auth requests


hunterlester:
SAFE Browser Version: head of `peruse` branch

theWebalyst:
That's interesting - are you sure that's SAFE Browser and not Peruse? I think this happened to me with the WebID test build of Peruse this week, but using an earlier official build of SAFE Browser this did not happen.

hunterlester:
@theWebalyst I should have specified `peruse` branch.

hunterlester:
Addressed by PR https://github.com/maidsafe/safe_browser/pull/280

PR #280, opened 2018-07-20
Issue #279 fix/notifier: successfully ignores auth request and allows further notifications

hunterlester:


app/components/Notifier/Notifier.jsx
```diff
@@ -36,14 +33,6 @@ export default class Notifier extends Component
         type        : 'alert'
     }
 
-    handleDismiss = () =>
```
joshuef: This doesn't seem right. We're removing a general 'ignore' function for the browser due to safe specific issues...

I'd say we still need this func to be available. And if we want something else to happen on ignore, we add that functionality it.

PR #284, opened 2018-07-26
 MAID-2946 fix/state: no longer producing unexpected UI after reloading safe-auth://home after logging out

hunterlester:
Also fixes network state updates for `auth-web-app` that weren't previously occurring.


hunterlester:
@joshuef @safesurfer Latest commit uses listener such that multiple instances of `safe-auth://home`, between windows as well, have synchronised state.

PR #286, opened 2018-07-30
Resolves #298 chore/ui: on did-fail-load, properly error and render fail message in UI

### LINKED ISSUE #298: prints message infinitely in console, when I use colon after `safe://` in the address bar
When I use colon/semicolon/question mark after `safe://` for ex, `safe://:play.ground` in the address bar, it infinitely prints the the following message until I close the tab.

```
[15:54:09.909] Setup up SAFE Dom API via @maidsafe/safe-node-app renderer.js:34 
[15:54:09.914] add Tab happening in reducer2renderer.js:34 
[15:54:09.917] add Tab happening in reducer renderer.js:34 
[15:54:09.921] add Tab happening in reducer renderer.js:34 
[15:54:10.378] Setup up SAFE Dom API via @maidsafe/safe-node-app renderer.js:34 
[15:54:10.384] add Tab happening in reducer 2renderer.js:34 
[15:54:10.387] add Tab happening in reducer renderer.js:34 
[15:54:10.392] add Tab happening in reducer renderer.js:34 
[15:54:10.851] Setup up SAFE Dom API via @maidsafe/safe-node-app
```

hunterlester:
Resolves #298

### DIFF
--- a/app/extensions/safe/test/e2e/safe.spec.js
+++ b/app/extensions/safe/test/e2e/safe.spec.js
@@ -106,26 +106,6 @@ describe( 'SAFE network webFetch operation', async () =>
     //     // expect( parsedUrl.protocol ).toBe( 'safe:' );
     // } );
 
-    it( 'validates URL before loading', async () =>
-    {
-        expect.assertions( 1 );
-
-        await setClientToMainBrowserWindow( app );
-        const { client } = await app;
-        const tabIndex = await newTab( app );
-        await delay(500)
-        await client.waitForExist( BROWSER_UI.ADDRESS_INPUT );
-
-        await navigateTo( app, 'safe://test-url.' );
-
-        const address = await client.getValue( BROWSER_UI.ADDRESS_INPUT );
-
-        await client.windowByIndex( tabIndex );
-        await delay( 2500 );
-        let text = await client.getText( 'body' );
-        expect( text ).toBe( 'Invalid URL: safe://test-url.');
-    } );
-
     if( ! isTestingPackagedApp )
     {
 
--- a/app/extensions/safe/test/e2e/safe.spec.js
+++ b/app/extensions/safe/test/e2e/safe.spec.js
@@ -106,6 +106,26 @@ describe( 'SAFE network webFetch operation', async () =>
     //     // expect( parsedUrl.protocol ).toBe( 'safe:' );
     // } );
 
+    it( 'validates URL before loading', async () =>
+    {
+        expect.assertions( 1 );
+
+        await setClientToMainBrowserWindow( app );
+        const { client } = await app;
+        const tabIndex = await newTab( app );
+        await delay(500)
+        await client.waitForExist( BROWSER_UI.ADDRESS_INPUT );
+
+        await navigateTo( app, 'safe://test-url.' );
+
+        const address = await client.getValue( BROWSER_UI.ADDRESS_INPUT );
+
+        await client.windowByIndex( tabIndex );
+        await delay( 2500 );
+        let text = await client.getText( 'body' );
+        expect( text ).toBe( 'Invalid URL: safe://test-url.');
+    } );
+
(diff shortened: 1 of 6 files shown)


safesurfer:
This issue is not only with colon, but also with other symbol characters like semicolon, tilde, backtick, ampersand, brackets, angle brackets, etc. 
But I tested with colon, it successfully shows the message `page load failed`.

joshuef:
I wonder if this is the right solution. While it seems to solve the problem. Maybe what we want is to handle invalid URLs better (like... not even passing them in here):

https://mathiasbynens.be/demo/url-regex

But that said, if they do sneak through (and as noted, no regex is perfect there...) we should _not_ be looping.

So what would that more generalised non-loop non-only-colon fix look like here?

app/extensions/safe/test/e2e/safe.spec.js
```diff
@@ -104,6 +103,26 @@ describe( 'SAFE network webFetch operation', async () =>
         expect( parsedUrl.protocol ).toBe( 'safe:' );
     } );
 
+    it( 'validates URL before loading', async () =>
```
joshuef: This should be a `peruse` level test, @hunterlester. The `safe` portion of the test isn't so needed as just any invalid URL (as it goes off of protocols declared in the package.json). So I'd suggest we move it there (and amend the test to be less 'safe')

joshuef:
@hunterlester this is blocking non-latin char urls at the moment:  http://見.香港/ registers as invalid (but should just be blocked...) eg. 

We should enable these chars (IMO), as well as other character sets.

Not sure if this would be a blocker or not though... thoughts @hunterlester @bochaco ?

bochaco:
I doubt we really want to validate the URL at all @joshuef @hunterlester , shouldn't we just make sure it's a `safe://` URL and let the underlying `webFetch` handle the rest? 
I wouldn't filter anything really, if the charset or characters is not matching with what can be found on the network then that's it. What's the issue in this approach?

joshuef:
@bochaco we need to prevent attempted loading of invalid urls. (well, 'need', is because it loops/fails atm, so that's an alternate angle).

I wouldn't bring `safe:` into it, as it can be handled at a peruse level (and we have some `http`/`localhost` type urls etc.

I agree in principle that we shouldn't be vetting what's in the url ideally though. Main point is for it not to die...

bochaco:
yes, correct, that's my view as well, it shouldn't break the browser and let any URLs with the protocols we want to support go thru and the underlying `webFetch` take care of it. In fact having the browser to filter URLs is incorrect to me as it's something about looking up the content on the network which is not a browser function

hunterlester:
@bochaco @joshuef I agree completely that we shouldn't be validating/limiting URL's like this.
This brings me back to the thought that validation should ONLY be happening on `did-fail-load` to determine the URL should attempted to be reloaded for it's it's simply an invalid URL that will never properly load?

hunterlester:
@joshuef @bochaco I've found that I wasn't asking the correct question when initially approaching this task. I should have been looking into why certain URL's were failing.
The regex implementation was not needed because URL's are validated upon `webview.loadURL(...)`.

This does bring up another question though. If someone wants to create a service name using colons or semicolons, which are valid parts of XOR names but then want to go to a website hosted by that service @ `safe://my:site.publicname` using Peruse, that URL will be considered invalid by `webview`'s `loadURL` function. How do we handle that?

bochaco:
If we have a limitation due to the webview that doesn't mean such a url is invalid or that we shouldn't allow, remember that it's still a possibility that apps use the `webFetch` directly. In an extreme case we can make sure the core applications that allow users to create public names disallow these characters, but I personally don't like it too much, at most I would just provide warnings but not disallow.

joshuef:
@hunterlester is there a possible non-regex solution here then? Can we detect the failed load, prevent the loop and show an error page without?

PR #306, opened 2018-08-27
MAID-3125: docs/license: update LICENSE details

pravarseth:
updated readme license text and license files to updated license details

### DIFF
--- a/LICENSE-BSD
+++ b/LICENSE-BSD
@@ -1,0 +1,28 @@
+BSD License
+
+Copyright (c) 2018 Blue Link Labs
+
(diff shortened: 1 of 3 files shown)


bochaco:
@pravarseth , I just checked and Beaker has only MIT license, can you please confirm if it's correct we add the BSD license to our fork of it and where you get this understanding from?

NickLambert:
Adding BSD makes this consistent with the rest of our licensing strategy @bochaco. There is very little difference between MIT and BSD but keeps thing simpler.

bochaco:
Ok thanks for the confirmation @NickLambert.

PR #318, opened 2018-09-04
Resolves #312 feat/error: creates error UI component

### LINKED ISSUE #312: Generate an error page for all type of errors when fetching data/pages

hunterlester:


hunterlester:
Tests consistently failing. Needs investigation.
What does `chrome not reachable` mean?

joshuef:
Do they still fail with latest master changes @hunterlester ?

Looks like tabs are being closed before tests are running or something...

hunterlester:
@joshuef What's happening here is that the state needs to be updated after a server `reply` is returned.
If I put this [dispatch](https://github.com/maidsafe/safe_browser/pull/318/files#diff-76922384c6ba85352a6a2acef42d9678R104) in a `setTimeout`, then all is well and tests pass.
However, I'm looking for something more reliable than a timeout.

I've been trying to use the Chrome debugger API to my advantage upon the `webview` `did-start-loading` event:
```
const attachDebugger = () =>
        {
            const debug = webview.getWebContents().debugger;
            debug.attach( '1.1' );
            debug.on( 'message', ( event, method, params ) =>
            {
                //logger.info( 'method: ', method );
                if ( method === 'Network.responseReceived' )
                {
                   // console.log('======================: ', params);
                    if ( params.response.url === webview.getURL() && params.response.status > 299 )
                    {
                        requestId = params.requestId;
                    }
                }
                if ( method === 'Network.loadingFinished' )
                {
                    if ( requestId )
                    {
                        debug.sendCommand( 'Network.getResponseBody', { requestId }, ( err, data ) =>
                        {
                            logger.info('Network.getResponseBody callback err: ', err);
                            logger.info('Network.getResponseBody callback data: ', data);
                        } );
                    }
                }
            } );
            debug.sendCommand( 'Network.enable' );
        };
```
To no avail as of yet.

app/components/PerusePages/Error/Error.jsx
```diff
@@ -0,0 +1,53 @@
...
+
+    static defaultProps =
+    {
+        error : { header: '', subHeader: '' }
+    }
+
+    render( )
+    {
+        const { error } = this.props;
+        const pageStyle = {
```
joshuef: @hunterlester can we align more with the other 'peruse pages', to use `Page` `Header` components etc (as makes sense), and keep the CSS in a separate file.

hunterlester: @joshuef Right, that's what I would prefer as well.
When I attempt to import nessie components into the `Error` component, I receive the following upon `yarn mock-dev`:
```
ReferenceError: navigator is not defined
    at C:\Users\guilf\safe\dev\browser\node_modules\nessie-ui\dist\webpack:\node_modules\codemirror\lib\codemirror.js:23:15
    at e (C:\Users\guilf\safe\dev\browser\node_modules\nessie-ui\dist\webpack:\node_modules\codemirror\lib\codemirror.js:18:5)
    at Object.module.exports.Object.defineProperty.value (C:\Users\guilf\safe\dev\browser\node_modules\nessie-ui\dist\webpack:\node_modules\codemirror\lib\codemirror.js:22:17)
    at n (C:\Users\guilf\safe\dev\browser\node_modules\nessie-ui\dist\webpack:\webpack\bootstrap:37:1)
    at e.defineMode.e.getMode.name (C:\Users\guilf\safe\dev\browser\node_modules\nessie-ui\dist\webpack:\node_modules\codemirror\mode\jsx\jsx.js:8:5)
    at Object.<anonymous> (C:\Users\guilf\safe\dev\browser\node_modules\nessie-ui\dist\webpack:\node_modules\codemirror\mode\jsx\jsx.js:28:14)
    at Object.module.exports.e.exports.default (C:\Users\guilf\safe\dev\browser\node_modules\nessie-ui\dist\webpack:\src\CodeEditor\codeEditor.css?9e69:2:19)
    at n (C:\Users\guilf\safe\dev\browser\node_modules\nessie-ui\dist\webpack:\webpack\bootstrap:37:1)
    at Object.module.exports.Object.defineProperty.value (C:\Users\guilf\safe\dev\browser\node_modules\nessie-ui\dist\webpack:\src\CodeEditor\index.jsx:133:18)
    at n (C:\Users\guilf\safe\dev\browser\node_modules\nessie-ui\dist\webpack:\webpack\bootstrap:37:1)
```
Must be something do to with importing resources in the Express server route, similar to how I wasn't able to utilize `ejs` as a view engine.

hunterlester:
Looks like this was the [only place](https://github.com/joshuef/peruse/commit/571084197b20ffb3c17f424355dac8a874e55c35#diff-afd99b3825a9184a99fa365152ec2029R13) where ejs was being utilized.

hunterlester:
"Why the extra commit?", you may be asking. I was unable to get the CI's to properly rebuild, even after deleting caches, etc.

__e2e__/peruse.spec.js
```diff
@@ -115,6 +115,45 @@ describe( 'main window', () =>
         expect( text ).toBe( 'Invalid URL: http://:invalid-url');
     } );
 
+    if ( nodeEnv === 'dev' )
```
joshuef: @hunterlester do we need to block `prod` from running these tests?

hunterlester: @joshuef This test fails if `NODE_ENV=prod` because it depends on the browser connecting to be able to make a web fetch. Intead of `example.com` returning `'Requested public name is not found'`, it will return `Setup Incomplete. Connection not available yet.`

Issue #320, opened 2018-09-05
No WebIds are displayed in the WebId switcher on Linux


bochaco:
When launching Peruse the auth pop-up for the peruse app itself doesn't show up, and therefore it's not fetching the list of WebIDs when clicking on the WebID switcher

PR #321, opened 2018-09-05
Add E2E tests for create account

joshuef:
And check if peruse is requesting to auth after account creation.


hunterlester:
Resolves #320

PR #323, opened 2018-09-05
chore(auth): able to paste into input via context menu option and value be analyzed for password strength

### LINKED ISSUE #281: Account setup secret and password are confused by pasted values
Operation System and Version: Windows 10 64bit
SAFE Browser Version: 0.10.2
Network: Testnet
Description: Pasting generated password gives "needs to be stronger" message
Steps to reproduce: Click on the account secret field. Paste a generated password value into the field. Error message "Account secret needs to be stronger."  This is even true if the password is a 100 character long random combination of upper lower numeric & symbol.   It appears that a single actual keystroke in the field solves the problem. This happens only on the account setup routine.  Once you have successfully set up the account, you can paste to login with no trouble.

hunterlester:
@joshuef @bochaco Note that this PR cannot be verified to resolve #281 until PR #315 is merged, after which the issue will exist as it currently only applies to SAFE Beaker Browser.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,11 @@
 # Peruse Change Log
 All notable changes to this project will be documented in this file.
 
+#### [Unreleased]
+
+### Fixed
+- Able to paste value into create_account input fields via context menu option and value be analyzed for password strength
+
 #### [0.7.0]
 ### Added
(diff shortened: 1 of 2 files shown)


hunterlester:
Resolves #281

PR #337, opened 2018-09-07
Resolves #336 chore/CI: only test packaged app

### LINKED ISSUE #336: Test only packaged app in CI

hunterlester:


package.json
```diff
@@ -11,8 +11,8 @@
     "test-peruse": "cross-env ELECTRON_RUN_AS_NODE=true NODE_ENV=test IS_UNPACKED=true BABEL_DISABLE_CACHE=1 node_modules/.bin/electron node_modules/jest/bin/jest.js __tests__ --notify",
     "test-watch": "yarn test peruse --watch",
     "test-exts": "cross-env ELECTRON_RUN_AS_NODE=true NODE_ENV=test IS_UNPACKED=true BABEL_DISABLE_CACHE=1 node_modules/.bin/electron node_modules/jest/bin/jest.js extensions --testPathIgnorePatterns=e2e --notify",
-    "test-e2e": "cross-env NODE_ENV=test SPECTRON_TEST=true IS_UNPACKED=true BABEL_DISABLE_CACHE=1 node runE2ETests.js e2e --runInBand",
-    "test-exts-e2e": "cross-env NODE_ENV=test SPECTRON_TEST=true IS_UNPACKED=true BABEL_DISABLE_CACHE=1 node runE2ETests.js exts-e2e --runInBand",
+    "test-e2e": "cross-env NODE_ENV=test SPECTRON_TEST=true IS_PACKED=true BABEL_DISABLE_CACHE=1 node runE2ETests.js e2e --runInBand",
```
joshuef: These e2e tests are running yarn and are not testing anything packaged, so this should not bee needed.

package.json
```diff
@@ -11,8 +11,8 @@
...
+    "test-exts-e2e": "cross-env NODE_ENV=test SPECTRON_TEST=true IS_PACKED=true BABEL_DISABLE_CACHE=1 node runE2ETests.js exts-e2e --runInBand",
```
joshuef: @hunterlester let's leave this line for running on dev machines.

We have `test-exts-e2e-prod` two lines below for this

joshuef:
Also, looking at commit hygiene: @hunterlester @bochaco I _think_, we dont want `resolves` as the first part of the commit title. 

I think we should have the `chore/ci` as the first part. And if we want resolves it should be part of the commit message body (on another line), after the title.

Thoughts?

bochaco:
Yes I think so, since that's the format we'll need for the automatic changelog generator

PR #366, opened 2018-09-13
fix(browser-window): adds frame to browser window options (#362)

### LINKED ISSUE #362: OSX open/close/minimise buttons, linux menu bar
With latest changes to window framing, we're missing some traditional buttonage on OSX, and it seems that latest versions of the ubuntu/gnome combo do net get a menu.

ubuntu/gnome is now the default desktop as opposed to unity in previous versions, so I'd say we need to treat gnome as first class here (and not unity).

hunterlester:
Resolves #362

### DIFF
--- a/app/openWindow.js
+++ b/app/openWindow.js
@@ -55,20 +55,14 @@ const openWindow = ( store ) =>
         y                 : newWindowPosition.y,
         width             : mainWindowState.width,
         height            : mainWindowState.height,
-        titleBarStyle     : 'hidden-inset',
-        'standard-window' : false,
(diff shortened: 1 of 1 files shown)


joshuef:
Looking fine on OSX.

joshuef:
Linux ubuntu unity going well. Setting up gnome now...

joshuef:
Okay, i cant get anything better with the menu (and gnome is sloooow). So I'm going to mark this as accepted 👍 

We can continue to look into gnome to streamline it in another issue.

PR #384, opened 2018-09-24
fix(state): wipes bookmarks and history upon logout, resolves #325

### LINKED ISSUE #325: History and Bookmarks not cleaned when logging out
In Peruse 0.7.0 I can logout and still view bookmarks & history. It is only cleaned when I restart Peruse.
Mock.
osx.

hunterlester:


app/reducers/tabs.js
```diff
@@ -443,13 +443,13 @@ export default function tabs( state: array = initialState, action )
         }
         case UI_TYPES.RESET_STORE :
```
joshuef: @hunterlester let's not change this behaviour by default. Otherwise `reset store` doesn't do what you expect.

__e2e__/lib/constants.js
```diff
@@ -1,6 +1,6 @@
 // TODO: Pull these from app constants, and add '.' programmatically.
 export const WAIT_FOR_EXIST_TIMEOUT = 20000;
-export const DEFAULT_TIMEOUT_INTERVAL = 35000;
+export const DEFAULT_TIMEOUT_INTERVAL = 55000;
```
joshuef: @hunterlester I'd tweak this value as it's pulled into each test instead of changing the overall default timeout

CHANGELOG.md
```diff
@@ -39,6 +39,7 @@ All notable changes to this project will be documented in this file. See [standa
 - Able to successfully auth when logging in on mock-dev on Windows
 - No longer hanging when logging in on Linux
 - Auths external apps on mock-dev and mock-prod
+- Bookmarks and history are wiped upon logout, without closing open tabs
```
joshuef: @hunterlester with the auto changelog we should not be adding to this list anymore, it'll be added automatically from commit messages.

Issue #380, opened 2018-09-18
Make sure package name is correct


joshuef:
We need to update naming for the repo to Safe Browser

This will need to update a couple of paths, as well as the electron process naming for OSX.

We should check that the browser loads properly, is packaged and named correctly and that auth uris etc load properly.

bochaco:
@hunterlester @joshuef , I see there is a `window.Peruse` object exposed which contains the `version`. 
I'm wondering if as part of this issue we can change it to be exposed at `window.browser`, `window.safebrowser`, or something else, and have the version and perhaps release name, e.g. `{ version: "0.7.0", release: "Peruse" }`

PR #391, opened 2018-10-02
chore(naming): Peruse is now SAFE Browser

hunterlester:


hunterlester:
Resolves #380

app/package.json
```diff
@@ -1,6 +1,6 @@
 {
-  "name": "peruse",
-  "productName": "Peruse",
+  "name": "safe-browser",
```
hunterlester: @bochaco @joshuef Should this be `safeBrowser` instead?
Otherwise we'd need to use `window["safe-browser"]`.

bochaco: I would agree, or would 'safe_browser' work and be preferable?

Issue #327, opened 2018-09-06
"Incorrect Password" authenticator error in Peruse 0.7.0 appears below Account Secret field (osx)


S-Coyle:
Should appear below the pwd field when it's the pwd that's incorrect

S-Coyle:
verified as fixed (issue seemed to have bypassed QA)

Issue #326, opened 2018-09-06
"entries_mismatch" authenticator error when mismatching secret/pwd confirmation


S-Coyle:
Could be handled better.

Peruse 0.7.0 mock.
osx.

PR #394, opened 2018-10-09
chore(UI): adds i18n into previous setup to improve error messages

hunterlester:


hunterlester:
Resolves #327 
Resolves #326
@S-Coyle When you have time, see what you think.

PR #398, opened 2018-10-16
feat(ui): enhances accessibility and i18n

### LINKED ISSUE #383: Enhance to comply with accessibility standards
https://github.com/electron/spectron#accessibility-testing
https://developer.mozilla.org/en-US/docs/Learn/Accessibility/HTML
https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA
https://reactjs.org/docs/accessibility.html

hunterlester:
Resolves #383 

Using [nvda](https://www.nvaccess.org/) screen reader to test browser.


hunterlester:
Resolves #465 

Reviving this old PR which solves some unexpected cursor behaviors.
It resolves #465 by NOT focusing the cursor upon loading authenticator into webview, so user can simply hit `tab` and it will focus as expected. This is the best solution as `webview` and `webContents` focusing is finnicky.

PR #403, opened 2018-10-30
Fix devtools warnings

### LINKED ISSUE #330: Multiple devtools warnings:
* "Hash history cannot PUSH the same path..."
* "Each child in an array or iterator should have a unique 'key' prop..." x2

Peruse-v0.7.0-rc.2 prod
osx

image

No noticeable adverse effect in my browsing as yet

shriram2301:
- fix router push PATH, and unique 'key' prop warnings
- fix window ID and webappkitregion warnings 

fixes #330

### DIFF
--- a/app/components/Browser/Browser.jsx
+++ b/app/components/Browser/Browser.jsx
@@ -50,6 +50,7 @@ class Browser extends Component
         super( props );
         this.state = {};
     }
+
     componentWillMount()
     {
 
--- a/app/components/Browser/Browser.jsx
+++ b/app/components/Browser/Browser.jsx
@@ -50,6 +50,15 @@ class Browser extends Component
         super( props );
         this.state = {};
     }
+    componentWillMount()
+    {
+
+        //jest/electron workaround as no remote in non-render process
+        const currentWebContentsId = remote ? remote.getCurrentWebContents().id : 1;
+
+        // this is mounted but its not show?
+        this.setState( { windowId: currentWebContentsId } );
+    }
 
     componentDidMount( )
     {
@@ -64,11 +73,7 @@ class Browser extends Component
 
         const theBrowser = this;
 
-        //jest/electron workaround as no remote in non-render process
-        const currentWebContentsId = remote ? remote.getCurrentWebContents().id : 1;
-
-        // this is mounted but its not show?
-        this.setState( { windowId: currentWebContentsId } );
+        
 
         if( !ipcRenderer ) return; //avoid for jest/Electron where we're not in renderer process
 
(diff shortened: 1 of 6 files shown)


app/components/Browser/Browser.jsx
```diff
@@ -50,6 +50,15 @@ class Browser extends Component
         super( props );
         this.state = {};
     }
+    componentWillMount()
```
bochaco: Cosmetic thing, perhaps keep an empty line to separate from previous function's body. Also perhaps remove the first empty line in this function's body.
```suggestion

    componentWillMount()
```

app/components/Browser/Browser.jsx
```diff
@@ -64,11 +73,7 @@ class Browser extends Component
 
         const theBrowser = this;
 
-        //jest/electron workaround as no remote in non-render process
-        const currentWebContentsId = remote ? remote.getCurrentWebContents().id : 1;
-
-        // this is mounted but its not show?
-        this.setState( { windowId: currentWebContentsId } );
+
```
bochaco: Again cosmetic, but important to keep the code clean and readable, perhaps remove one of the two empty lines?

app/extensions/safe/auth-web-app/components/app_list.js
```diff
@@ -104,10 +98,10 @@ export default class AppList extends Component {
 
   getNoAppsContainer() {
     return (
-      <div className="no-apps">
+      <div key={"noApps"} className="no-apps">
```
bochaco: Is the `key` really required in places like this one? assuming this is not part of a list of elements, if not, perhaps we shouldn't add them in this type of places to keep the code simpler and cleaner?

shriram2301: Noted, I will check and confirm in a bit.

joshuef:
Working well on OSX 👍

app/extensions/safe/auth-web-app/components/app_list.js
```diff
@@ -104,10 +98,10 @@ export default class AppList extends Component {
...
         <h3 className="no-apps-h">Looks like you haven&rsquo;t authorised any apps yet.</h3>
-        <div className="no-apps-img">{''}</div>
-        <div className="no-apps-down">
+        <div key={"noAppsIMG"} className="no-apps-img">{''}</div>
```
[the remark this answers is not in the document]
shriram2301: Noted. So, I ended up adding the keys for all the tags because prior to your commit for push Route warning, this was the only way to suppress the warning and finding the next place where the warning was present. I have  refactored it and will commit it after a round of testing.

PR #406, opened 2018-11-01
[WIP] fix: add icon to browserWindowConfig to display icon

shriram2301:


joshuef:
On linux dev mode has an icon, but the packaged version does not (ubuntu 18).

@shriram2301 this could be due to path resolution / folder structure in the package vs in dev mode

app/openWindow.js
```diff
@@ -47,6 +47,18 @@ const openWindow = ( store ) =>
         defaultHeight : 1024
     } );
 
+    let appIcon = path.join( __dirname, '../resources/safeicon.png' );
+        
+    if( process.platform === 'darwin' )
```
joshuef: We can drop the osx portion of this. It wont show as the icon is defined in Electron's `info.plist`. I wouldn't suggest we touch that, so that's okay.

shriram2301: I have dropped this portion as suggested. Please take a look and confirm if it works fine for you in linux and windows.

PR #409, opened 2018-11-07
chore(deps): upgrades electron to 3.0.9

hunterlester:


hunterlester:
I'm looking for leads as to why `webview` is crashing with black screen.
When pulling with PR for testing, replace `@maidsafe/safe-node-app` dep value with `git+https://github.com/hunterlester/safe_app_nodejs#305`.

PR #416, opened 2018-11-13
chore(notifications): reauthorisation renders original container permissions with different message

hunterlester:


app/extensions/safe/test/e2e/safe-auth.spec.js
```diff
@@ -179,6 +182,39 @@ describe( 'safe authenticator protocol', () =>
         } );
     }
 
+    it.only( 'can login, log history, logout, and clean up history', async ( ) =>
```
joshuef: A lingering `.only` @hunterlester

hunterlester: @joshuef Oh right, thank you. I didn't mean to push to this branch. I'm experimenting with this e2e test.

PR #419, opened 2018-11-13
fix: show devtools in --debug mode after pack

shriram2301:
show the Browser app devtools in --debug mode after pack.

### DIFF
--- a/app/menu.js
+++ b/app/menu.js
@@ -10,7 +10,7 @@ import {
 } from 'actions/tabs_actions';
 
 import { selectAddressBar } from 'actions/ui_actions';
-import { isHot } from 'appConstants';
+import { isHot,isRunningDebug } from 'appConstants';
 import { getLastClosedTab } from 'reducers/tabs';
 import logger from 'logger';
 import pkg from 'appPackage';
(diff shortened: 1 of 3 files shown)


joshuef:
FYI @shriram2301 I'm only seeing the `app.html` window opened on OSX, not the bg process window.

joshuef:
Also, @shriram2301 you'll need ot make sure the devtools are not opened when running spectron, as this breaks the e2e tests.

app/menu.js
```diff
@@ -32,7 +32,11 @@ export default class MenuBuilder
         {
             this.setupDevelopmentEnvironment();
         }
-
+        
+        if ( isRunningDebug )
```
joshuef: @shriram2301 here you need to check we're not using `spectron`.

Also, I wonder if here is the best place to be triggering devTools opening and not in `openWindow` `setupBgWindow` files for example?

shriram2301: in `openWindow` I believe the `menuBuilder` initializes an object of type `MenuBuilder` which is used to create the browser menus by taking in various arguments, this is the class exported by `menu`. I figured since all the logic for stuff like `isHot` is already being used there makes sense to add the stuff for debug in the same place. I will check for `spectron` just to be sure and run it as well. For the `bg` I will restrict all changes to only `setupBgWindow`

PR #435, opened 2018-11-28
Experiment tag to RHS of address bar

### LINKED ISSUE #428: Move experimental Tag to RHS of address bar
The experimental indicator should be on the RHS portion of the address bar.

joshuef:
Fixes #428


joshuef:
added into warn branch in the end

PR #448, opened 2018-12-03
fix(UI): when the app name is just - or _ characters render them in the authenticator as they are

bochaco:
Minor fix for an unlikely but possible case when the application's name is just a sequence of `-` and/or `_` characters. In such don't replace theses characters with white spaces. This was causing the whole authenticator page to be a blank screen.

To reproduce it you can authorise an app with the following info and then go to the authenticator tab to see a blank page:
```
window.safe.initialiseApp({
    id: 'net.maidsafe.test.webapp.id',
    name: '-',
    vendor: 'MaidSafe Ltd.'
  });
```


safesurfer:
QA verified, I'm no longer seeing blank auth page, when the app name is just - or _ .

PR #458, opened 2018-12-10
test(settingsMenu): add e2e tests for settings menu

### LINKED ISSUE #424: Add e2e Tests for Settings menu
We need e2e tests for checking the settings menu opens and is populated with history/bookmarks

shriram2301:
resolves #424

### DIFF
--- a/__e2e__/peruse.spec.js
+++ b/__e2e__/peruse.spec.js
@@ -38,7 +38,7 @@ describe( 'main window', () =>
 
     beforeEach( async () =>
     {
-      app = setupSpectronApp();
+        app = setupSpectronApp();
 
         await beforeAllTests(app)
     } );
@@ -150,8 +150,10 @@ describe( 'main window', () =>
         await client.click( BROWSER_UI.BACKWARDS );
         await client.pause( 4500 );
         await client.windowByIndex( tabIndex );
+        await delay( 4500 );
 
         const clientUrl = await client.getUrl();
+        await client.pause( 4500 );
         const parsedUrl = urlParse( clientUrl );
 
         expect( parsedUrl.host ).toBe( 'example.com' );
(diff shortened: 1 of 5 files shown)


__e2e__/peruse.spec.js
```diff
@@ -242,8 +244,84 @@ describe( 'main window', () =>
...
 
+        await setClientToMainBrowserWindow( app );
+        await delay( 4500 );
+        await client.click( BROWSER_UI.SETTINGS_MENU__BUTTON );
+        await delay( 4500 );
+        const settings = await client.waitForExist( BROWSER_UI.SETTINGS_MENU, WAIT_FOR_EXIST_TIMEOUT );
+        await delay( 2500 );
+        expect( settings ).toBeTruthy();
+    } );
+    it('can open settings menu and checks if Bookmarks,History,Toggle exist', async () =>
```
joshuef: Let's leave a space between the previous test and the next always, @shriram2301

__e2e__/peruse.spec.js
```diff
@@ -242,8 +244,84 @@ describe( 'main window', () =>
...
+    {
+        expect.assertions( 1 );
+        const { client } = app;
+        await delay( 4500 );
 
+        await setClientToMainBrowserWindow( app );
+        await delay( 4500 );
+        await client.click( BROWSER_UI.SETTINGS_MENU__BUTTON );
+        await delay( 4500 );
+        const settings = await client.waitForExist( BROWSER_UI.SETTINGS_MENU, WAIT_FOR_EXIST_TIMEOUT );
```
joshuef: similar to how you have `menuExists` above, lets make sure the variable naming that we're checking is clear.

So `settingsMenuIsShown` for eg here, (but also check throughout).

__e2e__/peruse.spec.js
```diff
@@ -242,8 +244,84 @@ describe( 'main window', () =>
...
+        await delay( 2500 );
+        expect( header ).toBe( 'Bookmarks' );
+    } );
+    it('can open settings menu and navigate to history', async () =>
+    {
+        expect.assertions( 1 );
+        const { client } = app;
+        await delay( 4500 );
+        await setClientToMainBrowserWindow( app );
+        await delay( 4500 );
```
joshuef: unless the CI is throwing errors, let's leave a smaller delay in general please, as `4500` is quite long, especially cumulatively.

__e2e__/peruse.spec.js
```diff
@@ -242,8 +244,84 @@ describe( 'main window', () =>
...
+        await client.click( BROWSER_UI.SETTINGS_MENU__BUTTON );
+        await delay( 4500 );
+        const settings = await client.waitForExist( BROWSER_UI.SETTINGS_MENU, WAIT_FOR_EXIST_TIMEOUT );
+        await delay( 2500 );
+        const bookmarks = await client.waitForExist( BROWSER_UI.SETTINGS_MENU__BOOKMARKS, WAIT_FOR_EXIST_TIMEOUT );
+        await delay( 2500 );
```
joshuef: do we need each of these delays?

joshuef:
@shriram2301 can you update the commit name as well to not use `feat` but use `test(<thing>): <description>`

__e2e__/peruse.spec.js
shriram2301: Sure thing, I've added spaces to all tests now.

__e2e__/peruse.spec.js
shriram2301: I have renamed the variable names appropriately.

__e2e__/peruse.spec.js
shriram2301: No we don't, I've tested out and kept the ones that we need.

__e2e__/peruse.spec.js
shriram2301: Noted I've used the delays in other tests as reference for these tests. But in general I agree that it should be as minimal as possible.

app/extensions/safe/components/wrapAddressBarButtonsRHS.js
```diff
@@ -94,10 +94,10 @@ const wrapAddressBarButtonsRHS = ( AddressBarButtons, extensionFunctionality = {
                     type="flex"
                     justify="space-between"
                     align="middle"
-                    className={ styles.toggleRow }
+                    className={ `${styles.toggleRow} ${CLASSES.SETTINGS_MENU__TOGGLE}` }
                 >
-                    <Col span={6} className={ styles.toggleText }>Toggle Experiments</Col>
-                    <Col span={6} offset={6}>
+                    <Col span={6} className={ `${styles.toggleText} ${CLASSES.SETTINGS_MENU__TOGGLE_TEXT}` }>Toggle Experiments</Col>
+                    <Col span={6} offset={6} classname={ `${styles.toggleText} ${CLASSES.SETTINGS_MENU__TOGGLE_BUTTON}` }>
```
joshuef: @shriram2301 two things here:

- This shouldn't have the `toggleText` class. It isn't text, and I'd wager that's not what the class is for at all. Please don't copy/paste indiscriminately, think about what you're adding.

- We should add the classes to the DOM element in question. Here you're adding them to the `Col` which is a layout tool. You should add your class to the `Switch` as that's what does the interaction (and if it were moved from without of this `Col`, it would then keep its class...)

Bonus point:

The toggle text classes are on a `Col`, it looks like as there's no wrapper on the text. Add a `span` around the text to achieve the same as the previous point above.

shriram2301: I missed this one out.  Since we were only planning on testing the `history and `bookmarks` I left it as is. This was in the initial commit when I wanted to add class names. Thanks for pointing it out. I didn't revisit it as all we needed to do was check in the elements just existed. I've changed it with changes you requested and tested it.

PR #463, opened 2018-12-11
chore(safe-node-app): Upgrade @maidsafe/safe-node-app dependency to v0.10.2

bochaco:


hunterlester:
Passes on Windows with the following changes: https://github.com/hunterlester/safe_browser/tree/ConfigJest-UpgradeEnzyme

PR #474, opened 2018-12-13
chore(safe-node-app): Upgrade @maidsafe/safe-node-app dependency to v0.10.3

### LINKED ISSUE #470: Getting error while creating WebID, but it is actually created and listed after page refresh
When I try to create a WebID using latest webid manager, it is showing an error saying `Error creating webid on the network` and If I switch to list tab from create tab it lists the id which was just throwing the error. 

Note: I used my old account which was created before rc2, for creating webid. Tested on Windows and Ubuntu cc: @bochaco 

image

image

bochaco:
Resolves #470

### DIFF
--- a/app/package.json
+++ b/app/package.json
@@ -17,7 +17,7 @@
   },
   "license": "MIT",
   "dependencies": {
-    "@maidsafe/safe-node-app": "0.10.2",
+    "@maidsafe/safe-node-app": "0.10.3",
     "cross-env": "5.1.3",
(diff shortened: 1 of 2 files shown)


bochaco:
I tested on Linux the scenario described in issue #470 and it's solved, the WebID is created and the publicName created is added to the _publicNames container without throwing an error in the WebID Mgr app. There are warnings being logged as expected since some entries cannot be decrypted by the application.

Issue #410, opened 2018-11-07
upgrade to latest electron version


hunterlester:
Attempted to upgrade electron to `3.0.8`.

Rebuild of `node-ffi` in `safe_app_nodejs` fails.

Several issues that `node-ffi@2.2.0` is not compatible with electron 3.0.8 which uses Node v10.
https://github.com/electron/electron-rebuild/issues/254

Supposedly, in `safe_app_nodejs` I should be able to change from `"ffi": "^2.2.0",` to `"ffi": "git+https://github.com/node-ffi/node-ffi.git",`, in order to use the fix that hasn't yet been published to npm but I haven't found success with it.

hunterlester:
[Electron v4.0](https://github.com/electron/electron/releases/tag/v4.0.0) released. Shall we have another go?

hunterlester:
We need to upgrade to Electron 4 to resolve an issue on OSX, see: https://github.com/maidsafe/safe_browser/issues/457

hunterlester:
Narrowed down the blank screen of death to be rooted in [Tab.jsx](https://github.com/maidsafe/safe_browser/blob/master/app/components/Tab/Tab.jsx).

hunterlester:
The only log that might be telling:
```
#
# Fatal error in , line 0
# Check failed: element->ToUint32(&number).
#
#
#
#FailureMessage Object: 00000001000FBAF8Backtrace:
17:40:03:992 Wrapping Address bar buttons LHS
17:40:04:004 Wrapping Address bar input
17:40:04:017 Wrapping Address bar buttons RHS
17:40:04:028 Wrapping Address bar input
        node::InternalCallbackScope::MarkAsFailed [0x00007FF6E474009B+76811]
        uv_fileno [0x00007FF6E56522C7+5341895]
        IsSandboxedProcess [0x00007FF6E4D84BD5+1476665]
        v8::internal::Isolate::random_number_generator [0x00007FF6E40254E2+38562]
        v8::internal::Isolate::random_number_generator [0x00007FF6E40243A1+34145]
        v8::internal::Isolate::random_number_generator [0x00007FF6E402423E+33790]
        v8::internal::Isolate::random_number_generator [0x00007FF6E4023367+29991]
        v8::internal::Isolate::random_number_generator [0x00007FF6E40225FB+26555]
        v8::internal::Isolate::random_number_generator [0x00007FF6E402391F+31455]
[...]

PR #477, opened 2018-12-15
[WIP] fix(reconnect): stabilises reconnection and reloading handling

### LINKED ISSUE #457: Problems with 'Reconnect Now'
This issue seems similar to PR #319, however there are some differences
- as mentioned in 319, it works with auth page, but it shows `Unable to navigate to site. Will automatically reload when network is connected...` message when I open a new tab to access a webpage.
- the new issue is intermittent and this also happens when I click `reconnect now`
- this time the reconnecting spinner spins infinitely
- and the browser becomes unresponsive, such as logout button doesn't work, unable to switch tabs, needs about 5 clicks to switch next tab. STR: try opening new tab and try to access some link, while it is loading click auth page tab

Note: I got this issue on Windows with prod version

hunterlester:
Resolves #457 

  - Fixed: Upon network disconnect, automated reconnect attempts are more
    stable
  - Added: Reconnect countdown UI
  - Added: On disconnect notification, user is given button to manually reconnect
  - Added: Test spec for new Notifier extension
  - Changed: Removes close tab/add tab behavior when load fails due to
    network disconnect. Replaced with reload behavior solely when network reconnects.

For QA:
Instead of waiting 2 minutes for network connection to be dropped upon simulating network disconnect, I switch from Ethernet connection to WIFI, which instantaneously triggers a network disconnect in the browser.

### DIFF
--- a/__tests__/components/Notifier.spec.js
+++ b/__tests__/components/Notifier.spec.js
@@ -1,30 +1,52 @@
 import React from 'react';
 import { mount, shallow } from 'enzyme';
 
-import Notifier from '../../app/components/Notifier';
+import Notifier from 'components/Notifier';
 import { Text } from 'nessie-ui';
+import { Provider } from 'react-redux';
+import configureStore from 'redux-mock-store';
 
-jest.mock('extensions', () => {
+const mockStore = configureStore();
 
-});
+// Some mocks to negate FFI and native libs we dont care about
+jest.mock( 'extensions/safe/ffi/refs/types', () => ( {} ) );
+jest.mock( 'extensions/safe/ffi/refs/constructors', () => ( {} ) );
+jest.mock( 'extensions/safe/ffi/refs/parsers', () => ( {} ) );
+
+jest.mock( 'ref-array', () => jest.fn() );
+//
+jest.mock( 'ffi', () => jest.fn() );
+jest.mock( 'extensions/safe/ffi/authenticator', () => jest.fn() );
+
+jest.mock( '@maidsafe/safe-node-app', () => jest.fn() );
+jest.mock( 'extensions/safe/actions/safeBrowserApplication_actions' );
 
 describe( 'Notifier', () =>
 {
     let wrapper;
     let instance;
     let props = {};
+    let store;
 
     beforeEach( () =>
     {
         props = {
-          isVisible          : false,
-          type               : 'alert',
-          acceptText         : 'Accept',
-          denyText           : 'Deny',
-          updateNotification : jest.fn(),
-          clearNotification  : jest.fn()
+            isVisible          : false,
+            type               : 'alert',
+            acceptText         : 'Accept',
+            denyText           : 'Deny',
+            updateNotification : jest.fn(),
+            clearNotification  : jest.fn(),
+            safeBrowserApp     : {
+                isConnecting : false
+            }
         };
-        wrapper = mount( <Notifier { ...props } /> );
+        store = mockStore( props );
+
+        wrapper = shallow(
+            <Provider store={ store } >
+                <Notifier { ...props } />
+            </Provider > ).dive();
         instance = wrapper.instance();
     } );
 
@@ -32,7 +54,7 @@ describe( 'Notifier', () =>
     {
         it( 'should have name Notifier', () =>
         {
-            expect( instance.constructor.name ).toBe( 'Notifier' );
+            expect( instance.constructor.name ).toMatch( 'Notifier' );
         } );
     } );
 
@@ -41,7 +63,10 @@ describe( 'Notifier', () =>
         beforeEach( () =>
         {
             props = { ...props, isVisible: true, text: 'notifier text' };
-            wrapper = mount( <Notifier { ...props } /> );
+            wrapper = mount(
+                <Provider store={ store } >
+                    <Notifier { ...props } />
+                </Provider > );
             instance = wrapper.instance();
         } );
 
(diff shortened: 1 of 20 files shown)


joshuef:
> Added: Reconnect key binding: CommandOrCtrl + ALT + R

I'm not sure we need this TBH. I mean... ideally you shouldn't be disconnected _soooo_ often that it's needed, right?

joshuef:
I'm still seeing some odd behaviour on OSX, lets have a HO and look at this later on today 👍 

eg: 

https://www.dropbox.com/s/d5gf7172igg7ku7/Screenshot%202019-01-03%2014.28.00.jpg?dl=0

joshuef:
@hunterlester just a follow up to the wee HO.

https://github.com/maidsafe/safe_browser/pull/477/files#diff-c1373fedfea725334ffc2413bc67f357L26

Here, we've lost auto reconnect (as the PR stands), which I think we need.

But it looks like how we had this set up was ripe for being hammered by reconnect calls which could have caused the bugginess we saw.

So we should try:

- [ ] Create a variable to track if a `timerIsRunning`, so a second call doesn't create a new timer when one is already running.
- [ ] Perhaps distinguish the timer from a manual click of the `attempt reconnect` button?
- [ ] Clear the timer once connection has been successfully established (and clear `timerIsRunning`also)

joshuef:
@hunterlester So this is now triggering reload when reconnected, which works for attempting to access a tab once it's known its disconnected.

BUT:

If i disconnect, then attempt to access a site (before the browser notes it's disconnected), the error page shows only `OK`. The reconnect notifier shows up, and reenabling connection, it _appears_ to be reloading (loading icon shows), however the page/correct error (no content found), is not shown. The page is blank.

All works as expected when attempting to open a tab once the browser is aware of being disconnected (correct error is shown, page reloads properly).

ALSO:

I noted, when disconnecting/reconnecting / clicking the button to reconnec tthat the countdown value started jumping around(eg: 10..4..8..12). So it seems that there are still _many_ counters running, each updating the state on the countdown shown to the user.

Last thing to note for now: Can you see if there's a nicer / cleaner way to present the reconnection indicator (or maybe we just remove the blue loading indicator etc... blue on red is a bit gross. (And maybe... does it need to be a red error notification, maybe info one suffices? (though im no sure that affects the colour situation there: P )

hunterlester:
- > If i disconnect, then attempt to access a site (before the browser notes it's disconnected), the error page shows only OK. The reconnect notifier shows up, and reenabling connection, it appears to be reloading (loading icon shows), however the page/correct error (no content found), is not shown. The page is blank.

    I'm working on reproducing this. Unable to thus far.

- Counter and loading presentation cleaned up. Instead of `error`, the notifier is an `info`, so looks a bit better for now until we switch over to `antd`.

- > (Should the notification be dismissible?)

    It's currently a default button, which I can change but I think it should be scoped for another task, since making it opt-in will affect other `addNotification` calls.

- > Perhaps const isDisconnected = state === SAFE.NETWORK_STATE.DISCONNECTED; needs to check if this is a fresh disconnect? Or one previously seen / notified / acted upon?
    
    As far as I can tell a 'Disconnect' network state event is only passed once when the network connection is dropped. Do you mean a subsequent `Disconnect` event after having reconnected and receiving a `Connected` network state event?

joshuef:
>  Do you mean a subsequent Disconnect event after having reconnected and receiving a Connected network state event?

Could be. Not sure why twas jumping around so. Will have a try w/ latest code and see what i can see

joshuef:
Info / button is looking okay yep!

I'm not seeing 'ok' anymore, nor the jumping counter 👍 

I am seeing that a window load that triggers awareness of disconnection still doesn't reload properly upon connection though.

hunterlester:
Able to reliably simulate network disconnect on Windows using Defender Firewall to create an outbound rule just for SAFE Browser.
Still unable to reproduce blank pages though @joshuef as well as that BG crash which occurs upon clicking `Reconnect` manually.
Need to dig deeper. Looking for a memory profiler that I can use to watch Electron's processes and what exactly is happening when 1. bg crash occurs and 2. when `webview` goes blank upon reconnect.

hunterlester:
@joshuef Got my hands on a Mac for testing purposes. Reproducing what you observed.
There are two primary issues:
1. When loading a site just before network disconnect is observed, then opening a blank tab, the loading tab triggers a `did-fail-load` webview event. Not able to reproduce this on Windows. This is an easy problem as I have to consider that either `webFetch` can error or Chromium webview can error, so I just have to handle both of those cases. Leading to the next problem.
2. I'm able to handle both Chromium and `webFetch` errors however and I'll observe a tab begin to reload as expected, HOWEVER, if that tab is not currently active, then the page will render blank, if it is active it will render the page as expected. That's where I'm leaving off for the day. Stumped on this one.
Does this make sense?
Anyway, I've pushed a commit, if you have any leads about what's going on here.

hunterlester:
Resolves #529

As-is, this PR improves tab behavior and reconnect behavior.
Reconnect edge case on OSX will be resolved upon updating electron to v4.

OSX edge-case:
```
I'm able to handle both Chromium and webFetch errors and observe a tab begin to reload as expected, HOWEVER, if that tab is not currently active, then the page will render blank, if it is active it will render the page as expected. That's where I'm leaving off for the day. Stumped on this one.
```
I'm paranoid now about regressions, so I'll be adding tests to prevent them and subject this PR to critical QA from @calumcraig and @safesurfer

hunterlester:
Blocked by #410

hunterlester:
Closing this PR because browser architecture has and will drastically change since it was originally opened.

PR #485, opened 2018-12-18
test: Add e2e test for `Mock` tag

### LINKED ISSUE #425: Add e2e test for `Mock` tag
We should check the `mock` tag exists in mock mode

shriram2301:
resolves #425


shriram2301:
-[ ] test if `Mock Network` tag is present or absent based on whether node env is `test` or `dev` or `prod`

joshuef:
Also `peruse/spec` is the wrong place for these tests, they are SAFE specific, so should live in the extensions.

app/constants.js
```diff
@@ -234,7 +234,8 @@ export const CLASSES = {
     SETTINGS_MENU__HISTORY       : 'js-settingsMenu_history',
     SETTINGS_MENU__TOGGLE        : 'js-settingsMenu_toggle',
     SETTINGS_MENU__TOGGLE_BUTTON : 'js-settingsMenu_toggleButton',
-    SETTINGS_MENU__TOGGLE_TEXT   : 'js-settingsMenu_toggleText'
+    SETTINGS_MENU__TOGGLE_TEXT   : 'js-settingsMenu_toggleText',
+    MOCK_TAG                     : 'js-MockTag'
```
joshuef: Please keep this inline with other class names (ie. camel case)

joshuef: and perhaps provide a more descriptive name

__e2e__/peruse.spec.js
```diff
@@ -340,6 +340,52 @@ describe( 'main window', () =>
...
+        await delay( 4500 );
+        await newTab( app );
+        await delay( 2500 );
+        if ( nodeEnv === 'dev')
+        {
+            expect( await client.waitForExist( BROWSER_UI.MOCK_TAG, WAIT_FOR_EXIST_TIMEOUT ) ).toBeTruthy();
+        }
+        if ( nodeEnv === 'prod')
+        {
+            expect( await client.isExisting( BROWSER_UI.MOCK_TAG, WAIT_FOR_EXIST_TIMEOUT, true ) ).toBeFalsy();
```
joshuef: Just running this text, in mock passes. Ie: this test doesn't appear to actually check what you want...

Why is it different to the above test? (ln 352?)

joshuef: These should really be two separate tests (running only if in that environment)

app/constants.js
```diff
@@ -38,6 +38,11 @@ if ( isRunningSpectronTestProcess && !isRunningSpectronTestProcessingPackagedApp
     shouldRunMockNetwork = true;
 }
 
+if ( isRunningSpectronTestProcess && !isRunningSpectronTestProcessingPackagedApp && allPassedArgs.includes( '--live' ) )
```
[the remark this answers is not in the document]
shriram2301: Yes, it is possible,I tried running it locally just now with this suggestion.I'll remove this `if` and reorder the `if` statements.

app/extensions/safe/test/e2e/safe-mock-live.spec.js
```diff
@@ -0,0 +1,85 @@
...
+    } );
+
+    it( 'checks for Mock Network Tag to be absent when --live parameter is passed', async () => 
+    {
+        expect.assertions ( 1 );
+        const { client } = app;
+        await delay( 4500 );
+        await newTab( app );
+        await delay( 2500 );
+        expect( await client.isExisting( BROWSER_UI.MOCK_TAG, WAIT_FOR_EXIST_TIMEOUT, true ) ).toBeFalsy();
```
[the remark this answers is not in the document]
shriram2301: oh ok I get it. The mock and live ones are disparate. I will change this.

PR #505, opened 2018-12-28
fix(tabs): focuses reopened last tab

### LINKED ISSUE #484: Reopen last tab doesn't focus on newly opened tab
**Steps to reproduce:**
1. Open SAFE browser with multiple (eg 5) tabs open.
2. Close one tab.
3. Select "Reopen last tab" from OR use CTRL + SHFT + T to open previously closed tab.

**Expected:**
Focus will change to newly reopened tab and bring it to the front of the browser window.

**Actual:**
Focus stays on currently open tab - reopened tab is in the "background" of the browser window. See attached document for screenshots:
https://docs.google.com/document/d/1jQUS65RJ0drXJ9elxnD9CsIdF0LiHBL-RXoiqTkCHbI/edit

hunterlester:
Resolves #484

### DIFF
--- a/app/reducers/tabs.js
+++ b/app/reducers/tabs.js
@@ -249,9 +249,10 @@ const reopenTab = ( state ) =>
     let { lastTab, lastTabIndex } = getLastClosedTab( state );
 
     lastTab = { ...lastTab, isClosed: false, closedTime: null };
-    const updatedState = [...state];
+    let updatedState = [...state];
 
     updatedState[lastTabIndex] = lastTab;
+    updatedState = setActiveTab( updatedState, { index: lastTabIndex } );
 
     return updatedState;
 };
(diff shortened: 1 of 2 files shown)


app/reducers/tabs.js
```diff
@@ -245,10 +245,11 @@ const reopenTab = ( state ) =>
 {
     let { lastTab, lastTabIndex } = getLastClosedTab( state );
 
-    lastTab = { ...lastTab, isClosed: false, closedTime: null };
-    const updatedState = [...state];
+    lastTab = { ...lastTab, isClosed: false, closedTime: null, isActiveTab: true };
```
joshuef: @hunterlester here I think we need to be calling `setActiveTab` with the `lastTabIndex` actually, (which will ensure the previous active tab is unset), and using that as the state.

(For me this isn't working on OSX).

I'd also suggest we add a jest test into the reducer for this case also 👍

hunterlester:
@joshuef Okay, see how that works now on OSX.

Issue #429, opened 2018-11-28
Cannot fetch content with a XOR-URL which typeTag is greater than 65535


bochaco:
The browser consider a XOR-URL invalid when the typeTag part is greater than 65535 e.g. `safe://hyfktcegr68th4og68y9eks99cie14qtw44zpmjdttqqujxr3bb5jnnip7r:160485` triggers error `Invalid URL: safe://hyfktcegr68th4og68y9eks99cie14qtw44zpmjdttqqujxr3bb5jnnip7r:160485`

joshuef:
So i've done some digging around, and this is for sure not even reaching any webRequest we can hook into.

So it looks, we file a bug on the basis that this isn't expected behaviour (as _technically_ port numbers can be larger based on the spec).

BUT, this behaviour exists in chrome (larger ports are just treated as search queries and no attempt is made to resolve them...), so I doubt very much that it'll get much traction.

SOOOOOOOOooooo...

How do we want to handle this? We can attempt an ugly workaround... catch the errors... check if its a BIG port number/XOR url, then fetch it, and server that from another URL, which we then _mask_ with modifying headers etc...

Or... something else?

hunterlester:
@bochaco @joshuef I'm confused, looking at the [TCP RFC, page 15](https://tools.ietf.org/html/rfc793), a port can only be a 16-bit value, right?

hunterlester:
Ah I see, port range not specified in URL spec: https://www.w3.org/Addressing/URL/url-spec.txt
It seems right though to assume a 16-bit port number as defined by TCP/IP and that we shouldn't conflict with it.

When encoding the XOR-URL, I think we need to change from using a colon to specify MData type tags to another character such as search query. `?`, such that `safe://hyfktcebyxbzht6j6o7fcwbbjh4o3chp16131q437t3kst5foer6ehzchny?150001`.

hunterlester:
The encoding ues multiformats such as multibase, should we also have a multiformats for our network data types so that, for example, the encoding specifies whether it represents an MData and should have a type tag or whether it's an IData and should not have a type tag?

joshuef:
I think the multiformats has already been covered in the current implementation @hunterlester . It's definitely included in the deps now. Though not sure if it's used for exactly what you're describing.

I'm all for an alternative char I think. `?`: could this be confused with query params by a webapp?  Maybe something still valid in a url, but not used as part of URL structs in general.

From the URL spec:

```
                        
[...]

PR #507, opened 2019-01-03
[POC] chore(url): fetch ports(xor-urls), greater than 65535

hunterlester:
Tracking #429 

@joshuef @bochaco See what you think of this WIP before I get too far in. The first commit is what's important to study. The second commit is my unsuccessful attempt to extend `Tab.jsx`, so that I could keep `safe` concerns separate from general Peruse concerns. I want Josh to take a look at that for suggestions.

`safe://hyfktcebyxbzht6j6o7fcwbbjh4o3chp16131q437t3kst5foer6ehzchny:150000` becomes `safe://hyfktcebyxbzht6j6o7fcwbbjh4o3chp16131q437t3kst5foer6ehzchny/?typeTag=150000` to make it through `webview` webRequest and then transformed back to be able to `webFetch`.

### DIFF
--- a/app/components/Tab/Tab.jsx
+++ b/app/components/Tab/Tab.jsx
@@ -669,14 +669,14 @@ For updates or to submit ideas and suggestions, visit https://github.com/maidsaf
     loadURL = async ( input ) =>
     {
         const { webview } = this;
-        const url = addTrailingSlashIfNeeded( input );
+        let url = addTrailingSlashIfNeeded( input );
         logger.silly( 'Webview: loading url:', url );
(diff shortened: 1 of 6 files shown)


joshuef:
@hunterlester very interesting! nice to see there's options, and that they're not too crazy either!

nice one. 👍 

This does make me think going for a different character as the delimiter will be cleaner though.

joshuef:
@hunterlester I'm closing this PR just to clean up the board. We'll still have this referenced in closed prs for the future 👍

PR #536, opened 2019-01-21
test(AddressBarInput): add extra unit tests for addressbar Input

### LINKED ISSUE #487: Add more unit tests for the address bar input component

### LINKED ISSUE #496: There are several issues in the WebID swticher that need to be enhanced/fixed
There are several issues in the WebID swticher/dropdown that need to be enhanced/fixed:

When the user is logged in, the item `Authorise to display your WebIds` is a link which even logs an error when clicked (`Uncaught TypeError: n is not a function`). That item should be just a label and the text should be improved as it's not clear what it means.

When the connecton to the network is slow, this brings some confusion to the user. The WebIDs are not displayed immediately when opening the dropdown, the `Authorise to display your WebIDs` link is shown for a few secs, then for another secs `No WebIds found` and `Updating WebIDs` items are shown, then after another few secs the webids are displayed, and this may even fail suposely due to some connection problems. We need to enhance this to have some visual indication of what's happenign in the back, perhaps a "loading" message/icon

All these issues were taken from the discussion taking place in the forum starting in this post: https://safenetforum.org/t/safe-browser-v0-11-0-release/26792/56

shriram2301:
resolves #487


__tests__/components/AddressBarInput.spec.js
```diff
@@ -67,4 +64,112 @@ describe( 'AddressBarInput', () =>
...
+            instance = wrapper.instance();
+            const input = wrapper.find( 'Input' );
+            input.simulate( 'focus' );
+            expect( props.onFocus ).toHaveBeenCalled();
+        } );
+
+        it( 'check if onKeyPress triggers', () =>
+        {
+            const input = wrapper.find( 'Input' );
+            expect( input.simulate( 'keyPress', { key: 'Enter', keyCode: 13, which: 13 } ) ).toBeTruthy();
```
joshuef: does this check `onKeyPress`? or just that the simulation was a success?

__tests__/components/AddressBarInput.spec.js
```diff
@@ -67,4 +64,112 @@ describe( 'AddressBarInput', () =>
...
+            input.simulate( 'keyPress', { key: 'Enter', keyCode: 13, which: 13 } );
+            expect( props.updateActiveTab ).toHaveBeenCalled();
+        } );
+
+        it( 'check if onChange triggers', () =>
+        {
+            const input = wrapper.find( 'Input' );
+            input.value = '123456';
+            input.simulate( 'change' );
+            expect( input.simulate( 'change' ) ).toBeTruthy();
```
joshuef: what's going on here with simulating change twice? 

I don't see what you're checking.

__tests__/components/AddressBarInput.spec.js
```diff
@@ -67,4 +64,112 @@ describe( 'AddressBarInput', () =>
...
+
+        it( 'check on onChange, if onSelect() is called', () =>
+        {
+            const input = wrapper.find( 'Input' );
+            input.value = '123456';
+            input.simulate( 'change' );
+            expect( props.onSelect ).toHaveBeenCalled();
+        } );
+    } );
+    describe( 'componentWillReceiveProps( nextProps )', () =>
```
joshuef: I don't think we need to check this as it's internal React functionality.

We can assume that's well tested.

shriram2301: alright then. I'll drop this from the tests.

__tests__/components/AddressBarInput.spec.js
joshuef: why are we testing that?

__tests__/components/AddressBarInput.spec.js
joshuef: is checking the simulation necessary? that's not our code. The simulation is enzyme code.

PR #537, opened 2019-01-21
fix(package): remove duplicate crust.config file in Windows package

### LINKED ISSUE #495: Ensure that duplicate crust.config doesn't exist in Windows package
If packaging on Windows, [this script](https://github.com/maidsafe/safe_browser/blob/master/post-pack.js#L55) copies `crust.config` into root release directory as case-sensitive file name. `safe-browser.crust.config` will have already been copied into that same directory.

Extra `crust.config` either needs to be deleted or even better would probably be to just [renameSync](https://nodejs.org/dist/latest-v10.x/docs/api/fs.html#fs_fs_renamesync_oldpath_newpath).

shriram2301:
resolves #495


hunterlester:
`crust.config` is copied into both the [resources](https://github.com/maidsafe/safe_browser/blob/master/package.json#L103) and [content](https://github.com/maidsafe/safe_browser/blob/master/package.json#L80) directories, which is the root package directory on Windows or Linux and is the `Contents` directory on OSX.

`post-pack.js` just copies the one from the `resources` directory, so it looks like we don't need `crust.config` to be copied into package contents. You'll have to confirm that.

`unlinkSync` in `post-pack.js` looks fine, just need to remove extra copy in `package.json`.

hunterlester:
To keep format for changelog generator, be sure commit message follows format, such as:
`fix(package): remove duplicate crust.config file in Windows package`

PR #540, opened 2019-01-21
fix(Blocked URL Handling): Blocked urls will now only send a tab back…

### LINKED ISSUE #539: Claim invitation causes multiple clearnet tabs to open.
**Steps to reproduce:**
1. Launch SAFE browser v0.11.1 RC2.
2. Click "Create account".
3. Click "Continue".
4. Click "Claim invitation".

**Expected:**
Clearnet browser is launched at invite.maidsafe.net

**Actual:**
Clearnet URL is opened but multiple instances of the clearnet browser is opened - usually around 6 but higher has been observed too.

image

joshuef:
…wards if it's possible to do such a thing.

Adds an if check to Tab.jsx error handling.
Adds tests for this portion of error handling.

fixes #539

### DIFF
--- a/__tests__/components/Tab.spec.js
+++ b/__tests__/components/Tab.spec.js
@@ -1,6 +1,5 @@
 import React from 'react';
 import { mount, render, shallow } from 'enzyme';
-
 import Tab from 'components/Tab';
 import { CLASSES } from 'appConstants';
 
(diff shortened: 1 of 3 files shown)


hunterlester:
When I execute the browser, `yarn mock-dev`, open new tab, use that tab to navigate to an `http` URI, the `BrowserWindow` goes blank:
blank-electron

At this point I can simply hit `CTRL+T` and the expected view appears again, although it doesn't open a new tab:
after-ctrl-t

I can go on as usual working with the browser.

The browser isn't actually crashing, either functionally or by monitoring electron processes, it's just this odd UI bug.

Here's something interesting. I happened to be experimenting with the `resizable` issue, removing `thickFrame` just before executing this branch, so the window open up with a black border between the frame and the `BrowserWindow`:
black-border

I don't know what it means what it was ubiquitous during my attempted electron 4 upgrade. I'm always suspicious that it means `crashed` but I don't see evidence yet.

joshuef:
Just FYI, @hunterlester what was happening was I'd used `updateTab` to close instead of `closeTab` cos I'm a dummy. The latter handles reactivating another tab, the former not. So there was no activeTab to display which is why empty UI (which is a situation where the window should close I guess [when it's happening appropriately]).

PR #543, opened 2019-01-22
 Fix Linting

### LINKED ISSUE #486: Fix lint issues and enable it so it's executed automatically with every push
We have to start working on enhancing the browser's code with what the linter reports, and we shall also enable this script to be executed autmatically when we push a commit to enforce the lint rules.

manavbp:
resolves #486


joshuef:
@manavbp if we're disabling `flow` let's also remove plugins / any other references to it. (I'm seeing some errors in lint output regarding flow)

joshuef:
@manavbp updating `eslintignore` to : 

```app/dist/
flow-typed/
app/dist/
app/main.js
app/webPreload.js
app/browserPreload.js
app/background.process.js
app/node_modules/
node_modules/
dll
*.map
release/
```

speeds things up a lot.

Issue #452, opened 2018-12-04
Shared MD authorisation request not rejected if MD doesn't exist


bochaco:
When sending an authorisation request for shared MD, there is no response if the targeted MD doesn't exist on the network.

Also, when clicking the `x` button in an authorisation popup a "rejected" auth response is sent back to the app instead of just ignoring it and not send any response at all:
![image](https://user-images.githubusercontent.com/442821/50484102-95578f80-09cd-11e9-82b2-c1e3d62c0b6e.png)

hunterlester:
Snippet to quickly test in API playground:

```
      const appInfo = {
        id: 'net.maidsafe.api_playground.webclient.9',
        name: 'SAFE web API playground',
        vendor: 'MaidSafe Ltd.',
        scope: null
      };

      try {
        let app = await safe.initialiseApp(appInfo);
        const authReqUri = await app.auth.genAuthUri();
        const authUri = await safe.authorise(authReqUri);
        app = await app.auth.loginFromUri(authUri);
        const hashedString = await app.crypto.sha3Hash('1010101010101');
        const permissions = [{
          typeTag: 15001,
          name: hashedString,
          perms: ['Insert']
        }];
        const shareMDataReqUri = await app.auth.genShareMDataUri(permissions);
        await safe.authorise(shareMDataReqUri);
      } catch (err) {
        return err;
      }
```

hunterlester:
Setting this aside to think about how best to resolve a few contributing issues that comprise this issue.
WIP: https://github.com/hunterlester/safe_browser/tree/452

hunterlester:
Setting `blocked` on this issue until I confirm with SCL where the root lies.

Issue can be observed when `auth_decode_ipc_msg` is called on a `ShareMDataReq`, composed of an `XorName` which doesn't exist on the network.
Expecting to receive `CoreError(No such data - CoreError::RoutingClientError -> NoSuchData)`, assigned to the `err` variable, which I successfully log here: https://github.com/maidsafe/safe_client_libs/blob/master/safe_authenticator/src/ffi/ipc.rs#L179
Expecting this value to be passed as the 3rd argument of `o_err` callback.
However, I observe a `null` value, received by browser's `safe_authenticator` implementation.
Looking at the behavior of the `call_result_cb!` macro, in this context it passes a default `ptr::null()` as the third callback argument: https://github.com/maidsafe/ffi_utils/blob/master/src/macros.rs#L70
[...]

PR #554, opened 2019-01-24
fix(auth): proper values returned on auth_decode_ipc_msg errors

### LINKED ISSUE #454: Automatically reject authorisation request which mockbit doesn't match
For the scenarios when the browser is running for mock network, and an application sends an authorisation request which is for live network (or the other way around), the authenticator should be rejecting the authorisation automatically and sending such a response to the application. Currently the application doesn't receive any type of response.

hunterlester:
- Test case for authenticator error -208 IncompatibleMockStatus
  - Test case for authenticator error -103 when decoding share MData
    request for non-existent MData

Resolves #452 
Resolves #454 

@joshuef @bochaco I'm unable to properly test 2 new cases, see [auth.spec](https://github.com/maidsafe/safe_browser/pull/554/files#diff-a003f29f7e2f9aeecfe6e3fbb39e3d2e), because they both output a `-1 ERR_ENCODE_DECODE_ERROR`. I need to think about what I need to do to properly include these tests.
Also, I've decided to simply show an error notification in browser. I don't think there is an easy way to respond to an external app with an error. The errors occur when attempting to decode the auth requests and decoding is needed to be able to then encode a response. I'll continue to think about a solution.

QA:
Easiest way to test this PR would be to:
- Run API playground
- Initialise app
- Pass object to `authorise` operation and run, where the whole func to be executed as auth should look like: 
```
async () => {
  
        const authReqUri = {
          uri: 'safe-auth:bAAAAAAEKDQ7DAAAAAAACOAAAAAAAAAAANZSXILTNMFUWI43BMZSS4YLQNFPXA3DBPFTXE33VNZSC453FMJRWY2LFNZ2C4OIAC4AAAAAAAAAAAU2BIZCSA53FMIQECUCJEBYGYYLZM5ZG65LOMQGQAAAAAAAAAACNMFUWIU3BMZSSATDUMQXACAQAAAAAAAAAAADQAAAAAAAAAAC7OB2WE3DJMMCAAAAAAAAAAAAAAAAAAAIAAAAAEAAAAABQAAAABQAAAAAAAAAAAX3QOVRGY2LDJZQW2ZLTAQAAAAAAAAAAAAAAAAAACAAAAABAAAAAAMAAAAAA'
      };
  
      // Sends autorisation request to authenticator
      try {
        authUri = await safe.authorise(authReqUri);
      } catch (err) {
        return err;
      }
      return `Auth URI returned. Next, back in auth section, pass authUri to loginFromUri.`;
    }
```
- Expect to see an error notification
- Same steps for [encodedNonExistentShareMDataReq](https://github.com/maidsafe/safe_browser/compare/master...hunterlester:454?expand=1#diff-a003f29f7e2f9aeecfe6e3fbb39e3d2eR28)

To QA with external app:
- You could use my repo and [set `forceUseMock` to `false` here](https://github.com/hunterlester/safe-app-base/blob/master/renderer.js#L58), in order to get a `-208` error.
- To test for error `-103`, uncomment [this block](https://github.com/hunterlester/safe-app-base/blob/master/renderer.js#L60-L68), then replace [authUri here](https://github.com/hunterlester/safe-app-base/blob/master/renderer.js#L69), with the `shareMDataReqUri` variable.

### DIFF
--- a/app/extensions/safe/ffi/authenticator.js
+++ b/app/extensions/safe/ffi/authenticator.js
@@ -670,29 +670,23 @@ class Authenticator extends SafeLib
                 reject
             );
 
-            const decodeReqErrorCb = this._pushCb(
-                ffi.Callback(
-                    types.Void,
-                    [ types.voidPointer, types.FfiResultPointer, types.CString ],
-                    ( userData, resultPtr ) =>
+            const decodeReqErrorCb = this._pushCb( ffi.Callback( types.Void,
+                [types.voidPointer, types.FfiResultPointer, types.CString], ( userData, resultPtr ) =>
+                {
+                    const result = resultPtr.deref();
+                    const error =
                     {
-                        const result = resultPtr.deref();
-                        if (
-                            !(
-                                this[_reqErrListener]
-                                && this[_reqErrListener].len() !== 0
-                            )
-                        )
-                        {
-                            return;
-                        }
-
-                        logger.info( 'Error in auth callback.', result );
-                        this[_reqErrListener].broadcast( JSON.stringify( result ) );
-                        reject( result );
+                        error_code  : result.error_code,
+                        description : result.description
+                    };
+                    if ( !( this[_reqErrListener] && this[_reqErrListener].len() !== 0 ) )
+                    {
+                        return;
                     }
-                )
-            );
+
+                    this[_reqErrListener].broadcast( JSON.stringify( error ) );
+                    return reject( error );
+                } ) );
             try
             {
                 this.safeLib.auth_decode_ipc_msg(
--- a/app/extensions/safe/ffi/ipc.js
+++ b/app/extensions/safe/ffi/ipc.js
@@ -236,7 +236,12 @@ class ReqQueue
             .catch( err =>
             {
                 // FIXME: if error occurs for unregistered client process next
-                self.req.error = err.message;
+                self.req.error = err;
+                if ( allAuthCallBacks[self.req.id] )
+                {
+                    allAuthCallBacks[self.req.id].reject( err );
+                    delete allAuthCallBacks[self.req.id];
+                }
 
                 // TODO/BOOKMARK: leaving off here. share MData req URI is causing error when used to call auth_decode_ipc_msg in authenticator.js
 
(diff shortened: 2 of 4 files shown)


joshuef:
> The errors occur when attempting to decode the auth requests and decoding is needed to be able to then encode a response. I'll continue to think about a solution.

Hmm, yeh given that situation, it seems like a notification is the best thing we can do 👍

hunterlester:
I've taken note of this [TODO](https://github.com/maidsafe/safe_browser/blob/5b52869465a4174674865e591b023af30a51f1b2/app/extensions/safe/test/auth/auth.spec.js#L410), however I do think it should be for a separate task as it effects code in Authenticator web app which requires some extra testing to prevent regression.

joshuef:
Hmm. @hunterlester following your instructions I don't see any notification pop up :( on OSX. Will check on linux now

joshuef:
I'm only seeing https://www.dropbox.com/s/d9ds89pafxqpb4w/Screenshot%202019-02-19%2012.00.39.jpg?dl=0 on both osx/linux, FYI @hunterlester

hunterlester:
@joshuef Ah, I gave bad instructions.
The following object should be passed to `authorise`:
```
const authReqUri = {
    uri: 'safe-auth:bAAAAAAEKDQ7DAAAAAAACOAAAAAAAAAAANZSXILTNMFUWI43BMZSS4YLQNFPXA3DBPFTXE33VNZSC453FMJRWY2LFNZ2C4OIAC4AAAAAAAAAAAU2BIZCSA53FMIQECUCJEBYGYYLZM5ZG65LOMQGQAAAAAAAAAACNMFUWIU3BMZSSATDUMQXACAQAAAAAAAAAAADQAAAAAAAAAAC7OB2WE3DJMMCAAAAAAAAAAAAAAAAAAAIAAAAAEAAAAABQAAAABQAAAAAAAAAAAX3QOVRGY2LDJZQW2ZLTAQAAAAAAAAAAAAAAAAAACAAAAABAAAAAAMAAAAAA'
};
```

joshuef:
aha, right. I'll update the original comment to include full code for this.

Looking good now on osx

joshuef:
This has been placed atop the #598, so merging that PR will close this one. (And that ideally will form initial commit after next release)

joshuef:
merged in with #598

safesurfer:
Tested with live browser and mock application which @hunterlester suggested, and getting the error `-208 IncompatibleMockStatus` while authorising, as expected. - first scenario
Tested the live app with live browser after uncommenting the block and replacing auth uri with `shareMDataReqUri `, which hunter mentioned, it throws `-103 error` as expected - second scenario

PR #571, opened 2019-02-01
Remove activeTab related logic as default behavior

### LINKED ISSUE #522: SAFE Browser - cycling through tabs causes extra tabs to open
**Steps to reproduce:**
1. Launch SAFE browser.
2. Click on 1+ SAFE links to open pages in new tabs.
3. _While pages are loading_ use CTRL+ TAB to move through the open tabs.

**Expected:**
Focus moves in L>R order and cycles through open tabs.
**Actual:**
Multiple new tabs are opened. It is easy to quickly have many new tabs which causes the browser to spend a long time loading. This only seems to happen while the pages are being loaded. Once the tabs have finished loading, the tabbing behaviour seems "stable" and no new tabs are loaded when CTRL + TAB is used.
Attaching video to illustrate this better than text description.

tab_issue

Full video can be found here: https://drive.google.com/file/d/1msi4p35y4iPgUdzSKF6cRsHVOy4otIEQ/view?usp=sharing

### LINKED ISSUE #567: Remove activeTab related logic as default behavior
Child task for #557

hunterlester:
Closes #567 

Resolves #522 

QA: 
- Cycling through other tabs while other tab(s) are loading shouldn't interrupt/delay loading
- Big refactor, so this PR should be on regression watch.

### DIFF
--- a/app/components/Browser/Browser.jsx
+++ b/app/components/Browser/Browser.jsx
@@ -24,8 +24,6 @@ class Browser extends Component
         selectAddressBar   : PropTypes.func.isRequired,
         deselectAddressBar : PropTypes.func.isRequired,
         blurAddressBar     : PropTypes.func.isRequired,
-        reloadPage         : PropTypes.func.isRequired,
-        pageLoaded         : PropTypes.func.isRequired,
         addTab             : PropTypes.func.isRequired,
         closeTab           : PropTypes.func.isRequired,
         closeActiveTab     : PropTypes.func.isRequired,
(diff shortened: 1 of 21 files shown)


hunterlester:
@joshuef On master branch there are `2807` lint errors. After running `lint-fix`, there are `512`.

app/extensions/safe/components/webIdDropdown/WebIdDropdown.js
```diff
@@ -36,12 +36,12 @@ export default class WebIdDropdown extends Component
...
 
-    handleIdClick = webId =>
+    handleIdClick = ( webId ) =>
     {
-        const { updateActiveTab, windowId, showWebIdDropdown } = this.props;
+        const { updateTab, windowId, showWebIdDropdown } = this.props;
         // also if only 1 webID? mark as defualt?
-        updateActiveTab( { windowId, webId } );
-    };
+        updateTab( { windowId, webId } );
```
joshuef: Will this work without a target tab id?

app/components/Browser/Browser.jsx
```diff
@@ -129,6 +129,11 @@ class Browser extends Component
                     console.info( 'unhandled command: ', type );
             }
         } );
+        const body = document.querySelector( 'body' );
+        const div = document.createElement( 'div' );
+        div.setAttribute( 'class', 'no_display' );
```
joshuef: @hunterlester why are we not doing this with react components?

app/extensions/safe/components/webIdDropdown/WebIdDropdown.js
joshuef: (it does, as we're still using activeTab in the reducer)

joshuef:
@hunterlester I'd say this doesn't wholly cover #567 as we're still guessing the intended tab from whatever is 'active' at that moment in the store.

Ideally, we'd be removing this step and _always_ sending a tabId with the action.

That said. There's a lot of improvements here, and it's already a big PR. So I'm opting to include and I'll make another issue to continue towards full #567 .

PR #572, opened 2019-02-04
fix(internal-pages): Disable refresh button for Bookmarks and History…

### LINKED ISSUE #520: History tab/Bookmark tab reloading issue
Tabs are reloading when you switch to it, after tried reloading either `Bookmark page` or `History` page.

STR:
1. Open Bookmark tab or History tab
2. Click Refresh button, it doesn't seem reloading the page.
3. Try to switch auth page or any other tabs except those two tabs ^^

Expected:
History Tab/Bookmark tab should be reloaded when we click Refresh button.

Actual:
It reloads the tab which you are switching to.

hunterlester:
Resolves #520 

QA:
Bookmarks and History pages are not composed of `webview` instances, thus they have no reload behavior. This PR disables refresh button for those pages.


hunterlester:
> this doesnt address the case of opening the history/bookmarks from shortcut keys / menus nor directly entering into the address bar.

@joshuef Is this still the case? I pushed a commit to handle it a bit differently.

joshuef:
@hunterlester CI failing with: 

```
 TypeError: Cannot read property 'includes' of null
      61 |             icon="reload"
      62 |             aria-label={I18n.t("aria.reload_page")}
    > 63 |             disabled={activeTab.isLoading || activeTabUrl.protocol.includes( PROTOCOLS.INTERNAL_PAGES )}
         |                                                                    ^
      64 |             onClick={handleRefresh}
      65 |           />
      66 |         </Col>

```

So i think we just need first check `protocol` exists

joshuef:
Still failing w/ linux ci :| @hunterlester

PR #575, opened 2019-02-06
Show_menu restricted to Current Window only

### LINKED ISSUE #562: Certain actions performed in one browser window, reflecting in all the other instances of the browser
issue-new

STR:

1. Open SAFE Browser (built from latest master after commit d214d74) and open some instances using `New Window` or `Ctrl + N`
2. Now try to click Menu button (vertical ellipsis) which expands the menu. This will be reflected on all other instances of the browser
3. whatever we open from that menu will also be reflected on all other instances

for ex: opening a History page in one window, opens it on all other windows.

manavbp:
Resolves #562 

Steps to test the resolution.
For example:
```
QA:
The easiest way to test this PR would be to:
-Open a new window
-If upon showing the menu, it doesn't show the menu in all windows of the browser
-If upon reloading the active tab in one window, does not cause the active tabs in other windows to reload as well
-If upon opening history or bookmarks in one window, does not open history and bookmarks in other windows of the browser as well
```

### DIFF
--- a/app/actions/ui_actions.js
+++ b/app/actions/ui_actions.js
@@ -23,13 +23,42 @@ export const {
     pageLoaded,
     focusWebview
 } = createActions(
-    TYPES.SHOW_SETTINGS_MENU,
-    TYPES.HIDE_SETTINGS_MENU,
-    TYPES.BLUR_ADDRESS_BAR,
-    TYPES.DESELECT_ADDRESS_BAR,
-    TYPES.SELECT_ADDRESS_BAR,
-    TYPES.RESET_STORE,
-    TYPES.RELOAD_PAGE,
-    TYPES.PAGE_LOADED,
-    TYPES.FOCUS_WEBVIEW
+    {
+        SHOW_SETTINGS_MENU : [
+            undefined,
+            () => ( { scope: 'local' } ) // meta
+        ],
+        HIDE_SETTINGS_MENU : [
+            undefined,
+            () => ( { scope: 'local' } ) // meta
+        ],
+        BLUR_ADDRESS_BAR : [
+            undefined,
+            () => ( { scope: 'local' } ) // meta
+        ],
+        DESELECT_ADDRESS_BAR : [
+            undefined,
+            () => ( { scope: 'local' } ) // meta
+        ],
+        SELECT_ADDRESS_BAR : [
+            undefined,
+            () => ( { scope: 'local' } ) // meta
+        ],
+        RESET_STORE : [
+            undefined,
+            () => ( { scope: 'local' } ) // meta
+        ],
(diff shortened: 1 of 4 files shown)


app/components/AddressBar/AddressBar.jsx
```diff
@@ -74,9 +75,8 @@ export default class AddressBar extends Component
 
         const addATab = ( tab ) =>
         {
-            addTab( { url: `safe-browser://${tab}`, isActiveTab: true } );
+            addTab( { url: `safe-browser://${tab}`, isActiveTab: true, windowId: remote.getCurrentWindow().webContents.id } );
```
joshuef: @manavbp , I thin we have windowId in the parent `Browser` component, so ideally we'd just pass that down and avoid having to import electron.remote here.

app/actions/ui_actions.js
```diff
@@ -23,13 +23,42 @@ export const {
...
-    TYPES.BLUR_ADDRESS_BAR,
-    TYPES.DESELECT_ADDRESS_BAR,
-    TYPES.SELECT_ADDRESS_BAR,
-    TYPES.RESET_STORE,
-    TYPES.RELOAD_PAGE,
-    TYPES.PAGE_LOADED,
-    TYPES.FOCUS_WEBVIEW
+    {
+        SHOW_SETTINGS_MENU : [
+            undefined,
```
joshuef: @manavbp I'm not sure if the payload creator can be `undefined` here... Can you check w/r/t https://redux-actions.js.org/api/createaction#createactionsactionmap

manavbp: createAction(type, payloadCreator, MetaCreator)
payloadCreator must be a function, undefined, or null. If payloadCreator is undefined or null, the identity function is used.
https://redux-actions.js.org/api/createaction#createactiontype-payloadcreator

manavbp:
@joshuef will take a look at the issue on OSX

app/components/AddressBar/AddressBar.jsx
```diff
@@ -65,7 +65,7 @@ export default class AddressBar extends Component {
         const { addTab } = this.props;
 
         const addATab = tab => {
-            addTab({ url: `safe-browser://${tab}`, isActiveTab: true });
+            addTab( { url: `safe-browser://${tab}`, isActiveTab: true, windowId: this.props.windowId } );
```
joshuef: @manavbp when grabbing props, we tend to use destructuring at the top of the func (see how we grab `addTab` above). This keeps code cleaner and is marginally more optimised.

joshuef:
This is working for me on osx @manavbp . Just that small code quality tweak and we're good 👍

PR #581, opened 2019-02-09
fix(CI): Add missing WAIT for some auth CI drivers

### LINKED ISSUE #565: Improve CI times...
Are there tests we can drop? Do we need to run e2e on both dev/master?

What could we reasonably do reduce CI burden without sacrificing stability?

joshuef:
Resolves #565 (after other fixes) by reducing failed CI runs


joshuef:
To check: CI passing consistently: https://travis-ci.org/joshuef/safe_browser/builds/490882433 as well as the PR tests

PR #588, opened 2019-02-12
fix top padding based on OS specification

### LINKED ISSUE #586: [OSX] Leave padding top left for osx
## Prerequisites

Use MacOS

## Expected Behavior

Close/minimize icons should not overlap any UI.

## Current Behavior

https://www.dropbox.com/s/t4buvky59m27icq/Screenshot%202019-02-11%2014.43.56.jpg?dl=0

## Possible Solution

Use constant OS detection to add a class to tab bar to add more padding.

manavbp:
Resolves #586 
```
QA:
The easiest way to test this PR would be to:
-Initialize the browser on Mac OS
-And check if close/minimize icons do not overlap any UI.
(Similarly for windows)
```

### DIFF
--- a/app/components/TabBar/TabBar.jsx
+++ b/app/components/TabBar/TabBar.jsx
@@ -1,7 +1,8 @@
-// @flow
+/* eslint-disable */
 import { remote } from 'electron';
 import url from 'url';
 import React, { Component } from 'react';
+import ReactDOM from 'react-dom';
 import PropTypes from 'prop-types';
 import styles from './tabBar.css';
 import MdClose from 'react-icons/lib/md/close';
@@ -144,10 +145,9 @@ export default class TabBar extends Component {
             );
         });
     };
-
     render() {
         return (
-            <div className={styles.container}>
+            <div className= {[styles.container , (process.platform === 'darwin' ? styles.containerMac : "") ].join(' ')}>
                 <div className={styles.tabBar}>
                     {this.getTabs()}
                     <div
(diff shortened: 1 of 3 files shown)


app/components/TabBar/TabBar.jsx
```diff
@@ -147,7 +148,7 @@ export default class TabBar extends Component {
 
     render() {
         return (
-            <div className={styles.container}>
+            <div className={ navigator.appVersion.indexOf("Mac")!=-1? styles.containerMac : styles.container}>
```
joshuef: We have the OS determined in the global constants, lets use this if we can 👍

app/components/TabBar/tabBar.css
```diff
@@ -3,14 +3,28 @@
...
+    padding-right: 10px;
     background: rgba( 207, 208, 209, 1);
     z-index: 4;
     font-size: 1.4rem;
     -webkit-app-region: drag;
 }
 
+.containerMac
+{
+    flex: 0;
```
joshuef: Here, I _think_ the only thing needed is `padding-left: 100px`, the rest should not need to be defined if this is just an extra class added TO the current tabBar. (no top padding is needed to be honest).

app/components/TabBar/TabBar.jsx
joshuef: So aye, here should be both `styles.container` and IF MAC `styles.containerMac`

PR #618, opened 2019-02-28
chore(restructure):Restructure remaining js files to ts

manavbp:
Restructure remaining js files to TS

QA:-
1)Test Authenticator if it works as expected


safesurfer:
QA verified, authenticator is working fine.

Issue #620, opened 2019-03-01
WebID dropdown not working


JPL1:
Browser 0.11.2 Debian Linux and Windows 10

The dropdown says 'no WebIDs found' even though I have created some. Possibly related to #320

![web-ids](https://user-images.githubusercontent.com/12783198/53647515-a1fa9d00-3c35-11e9-8d33-0bb394881729.png)

theWebalyst:
I have this issue on Ubuntu 18.10 (SAFE Browser 0.11.2 dev build).

Any news / workarounds? I need this to develop stuff! Cheers.

calumcraig:
Apologies for not responding to this issue previously. This is a know bug which was reported previousy in the forum:
https://safenetforum.org/t/webids-not-showing-in-browser/27499
The devs are currently looking at it and a fix is in progress - not sure when they'll be able to deliver it at the moment but as soon as poss.

bochaco:
As a workaround, for me it works by first enabling experimental APIs and then loging in the Authenticator. 
Although I know that doesn't work for others, but worth a try.

theWebalyst:
Thanks folks.

@bochaco unfortunately that workaround doesn't make any difference for me but I can fudge the code for now, so not urgent for me.

joshuef:
I think I see the problem here with some post-refactor code.

Should have a commit for you to check out shortly @calumcraig !

calumcraig:
Testing on v0.11.3 RC. This issue has been resolved - clicking on the WedID Manager icon in the address bar now gives us a drop down menu showing the existing WebIDs plus an option to open the Manager app:
![image](https://user-images.githubusercontent.com/44327924/54443999-c5dfd780-4739-11e9-9ab2-0b3e533477ff.png)
Tested on Linux Mint 19 Cinnamon.

PR #629, opened 2019-03-13
fix(webId): Calls to get webIds fixed.

joshuef:
QA: fixes https://github.com/maidsafe/safe_browser/issues/620

WebId Dropdown should populate (after a short delay fetching)

### DIFF
--- a/app/extensions/safe/actions/safeBrowserApplication_actions.ts
+++ b/app/extensions/safe/actions/safeBrowserApplication_actions.ts
@@ -119,10 +119,10 @@ const getWebIds = async () =>
 
     let webIds = [];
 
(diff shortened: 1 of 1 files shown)


manavbp:
The web ids were populating for both mock & prod. But in prod after you log in, authenticator remains at the login page and the applist page doesnt show up. Instead, it remains at the login page, no matter how many times the page is refreshed

joshuef:
Aha, yeh I was suspecting something similar from what I've seen.

@manavbp reported some things, and @bochaco too. 

But my linux/osx tests are good, and this code literally just throws errors as it stands. So I think we should get this in, and look to improve experimental mode / webids in their separate issues 👍

PR #635, opened 2019-03-19
refactor(notifications): deprecates Notifier component in favor of antd

### LINKED ISSUE #619: Don't open HTTP(s) links by default
## Currently

Any HTTP/S request is blocked within the SAFE Browser, but opens your default browser automagically.

This may well be abused to deanonimize users.

## Wanted

At the base level: Not to open HTTP/S links automatically.

Show a notification w/ option to open a link? (We'd need this while we have the invite server eg.)

## Other considerations,

We should inform the user that there's a request. And of the (potential) dangers of a clearnet request.

We may want to consider:

- [ ] Blocking more than X notifications / requests from a page?
- [ ] Wording for any notifications/alerts
- [ ] Allow opening of links by default if a user initiated the process? (eg by clicking a link)
- [ ] A setting to block all HTTP/S notifications altogether. (Should that be default)
- [ ] A setting to _allow_ HTTP/S requests by default
- [ ] Do we allow other protocols to be opened by their default handler? Is that a separate notification check?
- [ ] Enable a whitelist of sites to open automatically? (And management of said list)
- [ ] .... more things...

@JimCollinson @shankar2105 FYI.

hunterlester:
- Changed: HTTP requests are no longer automatically forwarded to
    default browser
  - Added: Initial HTTP allowed-list containing 'https://invite.maidsafe.net'
  - Removed: Notifier component
  - Changed: clearNotification operations may be passed notification ID,
    otherwise defaults to shift notification array
  - Added: Static Notification type declaration
  - Changed: Updates relevant tests which cover notifications

Resolves #619 

QA:
- HTTP requests either via address bar or via Fetch API should NOT forward to system's default browser. Instead, notification should appear, providing user with option to copy to clipboard.
- One exception is that `https://invite.maidsafe.net` has been added to an allowed list of acceptable HTTP requests, so that URL should be forwarded to default browser.
- Make a non-allowed HTTP request such as: `https://www.rust-lang.org` and observe notification
- Make an `index.html` which fetches non-allowed resources and open in browser, such as:
```html
<html>
    <body>
        <img src="https://i2.wp.com/www.foodrepublic.com/wp-content/uploads/2013/09/glamorous-assorted-apples-pictures-hd-wallpaper-hd-pictures.jpg"> 
        <img src="https://img1.cookinglight.timeinc.net/sites/default/files/styles/4_3_horizontal_-_1200x900/public/image/2017/11/main/satsuma-oranges-getty-1711w.jpg"> 
    </body>
</html>
```
Notifications will NOT appear for these two HTTP resources. Two errors will be logged to the console.


joshuef:
@hunterlester I think we need to fire a notification for _each_ request to enable copying (as there could be many links / reqs rapidly... and we dont want to override clipboard without user opt in).

Then in _this_ notification, they can copy the link, or we provide a button to initiate copy eg.

Otherwise this is a decent first implementation.

Only other thing to add is maybe to override `window.fetch`, to return errors when attempting to fetch HTTP/S content so webapp devs can be more aware of what's going on?

hunterlester:
@joshuef See updated body above. See how you like the style and behavior now.

joshuef:
So I think this is great!

I dont think we need to notify for `fetch` requests, just return an error/ console log or so in the DOM.

(An app could fire many many requests).

For addressbar / clicking links etc, this is ideal.

Also the notification is soooo much nicer.

Is this applied across the board for notifications? If not, I think we should otherwise this'd be jarring.

hunterlester:
@joshuef 
`click to copy should have a pointer styles (and maybe be a button?)`
`click to copy isnt working for me on osx`
I'm removing the option to copy to clipboard from this PR, in favor of a separate PR.
My implementation hard-coded the clipboard behavioir, `onClick`, in the `handleNotifications` operation. I was attempting to correct this by passing an arrow function as a part of the notification state, which is not supported as it cannot be properly serialized.

`can we get e2e testcafe tests for the presence of the notifier`
Currently unable to trigger notification with testcafe. Not sure why as it's not dependent on `webview`.
However, this is currently covered by existing spectron tests for now.

`we need unit tests for handleNotification`
Initial existential test.
Not sure how else to unit test `handleNotifications` as it produce side-effect and no return value.
The side effect is of calling `notification.<type>` and I'm not sure how to mock it.
I tried something like:
```
import { notification } from 'antd';
jest.mock( 'antd');

    it( 'opens antd notification', () => {
        const prevBrowserProps = {notifications: []};
        const currBrowserProps = {notifications: [ { id: "ie93dk203", type: "error" } ]};
        handleNotifications(prevBrowserProps, currBrowserProps);
        expect( notification.error ).toHaveBeenCalled();
    } );
```
But jest is complaining that the value passed to `expect` is not a mocked function.

`we need unit tests for ReactNodeToElement`
Done

joshuef:
> I'm removing the option to copy to clipboard from this PR, in favor of a separate PR

Fair enough 👍 So the scope of this is just to always block + show a notification (and that all notifications are in the new style)

> Not sure why as it's not dependent on webview.

Actually this makes sense. The notification is triggered when the webview makes a request. Not when it's entered into the address bar.

I'll pull and have a look at `handleNotification` now.

joshuef:
@hunter , this works:

```js
import handleNotifications from '$Utils/handleNotificiations';
import { notification } from 'antd';
import { reactNodeToElement } from '$Utils/reactNodeToElement';

jest.mock( '$Utils/reactNodeToElement', () => ( {
    reactNodeToElement: jest.fn()
} ) );

jest.mock( 'antd/lib/notification', () => ( {
    error : jest.fn()
} ) );


describe( 'handleNotifications', () => {
    it( 'should exist', () => {
        expect( handleNotifications ).not.toBeNull();
    } );

    it( 'opens antd notification', () => {
        const prevBrowserProps = {notifications: []};
        const currBrowserProps = {notifications: [ { id: "ie93dk203", type: "error" } ]};
        handleNotifications( prevBrowserProps, currBrowserProps );
        expect( notification.error ).toHaveBeenCalled();
        expect( reactNodeToElement ).not.toHaveBeenCalled();
    } );
} );

```

If we run `expect( notification.error ).not.toHaveBeenCalled();` you can see all the args it was called with. From there we can check the correct things are being passed too.

We can also test that it called `reactNodeToElement` when something not reactNode is passed eg.

That should be sufficient I think here.

.vscode/settings.json
```diff
@@ -16,5 +16,10 @@
             "language": "typescriptreact",
             "autoFix": true
         }
-    ]
+    ],
+    "files.exclude": {
+        "**/lib_zips": true,
```
joshuef: Ah this shoulnd't exist actually. `lib_zips` has been accidentally (by me I think) when testing some methods of managing safe_app native libs downloading. So we can remove this line (and if the folder is still in the repo, remove it too)

app/extensions/safe/__testcafe__/e2e/safe.spec.ts
```diff
@@ -0,0 +1,25 @@
+import { ClientFunction, Selector } from 'testcafe';
+import { ReactSelector, waitForReact } from 'testcafe-react-selectors';
+import {
+  getPageUrl,
+  getPageTitle,
+  navigateTo,
+  resetStore
+} from '../../../../../__testcafe__/helpers';
```
joshuef: lets maybe add an alias for testcafe?

hunterlester:
@joshuef Updated. Thank you for your help.

Issue #636, opened 2019-03-19
Performance issue when 1+ browser window open - MacOS


calumcraig:
**Steps to reproduce:**
1. Launch browser and open more than one browser window.
2. Open more than one tab in each window and load some pages.

**Expected:**
Performance should not be adversely affected.
**Actual:**
Performance seriously impacted - browser process seen at around 100% of processor. Browser becomes extremely unresponsive, clicking on menu, changing tab etc doesn't work / takes excessively long.
<img width="1427" alt="Screenshot 2019-03-19 at 10 55 26" src="https://user-images.githubusercontent.com/44327924/54602866-ba8ef380-4a3a-11e9-9d35-32a55dad120f.png">
Closing one browser window (after a delay of several seconds) will alleviate the issue and performance returns to normal:
<img width="1418" alt="Screenshot 2019-03-19 at 11 02 14" src="https://user-images.githubusercontent.com/44327924/54602873-c11d6b00-4a3a-11e9-8990-4159bb97892f.png">

calumcraig:
Above behaviour observed on MacOS.

The same scenario in Linux Mint shows no effect on performance:
![image](https://user-images.githubusercontent.com/44327924/54603519-6d138600-4a3c-11e9-89ee-a25f9896dfb9.png)

safesurfer:
I've tried opening 5+ browser windows and many tabs in one window, as you said it went to 68% of CPU from 10% for few seconds, but it dropped to 8% and still it is in 8-10%. I didn't close any of the browser windows. still they are opened

joshuef:
this was fixed in #766

calumcraig:
Tested latest commit on `dev` branch: `865e1b3c (HEAD -> dev, origin/dev, origin/HEAD) chore(deps-dev): bump eslint-plugin-jest from 22.5.1 to 22.6.4`

Test machine - QA Mac:
``Model Name:	iMac
  Model Identifier:	iMac11,3
  Processor Name:	Intel Core i3
  Processor Speed:	3.2 GHz
  Number of Processors:	1
  Total Number of Cores:	2
  L2 Cache (per Core):	256 KB
  L3 Cache:	4 MB
  Memory:	12 GB
  Boot ROM Version:	96.0.0.0.0
  SMC Version (system):	1.59f2
  Serial Number (system):	W80520B5DB6
  Hardware UUID:	33D60AE5-DFAC-58BB-B43D-576149FE4F32```

[...]

Issue #599, opened 2019-02-20
"Open link in new tab" opens at end of currently open tabs.


calumcraig:
**Steps to reproduce:**
1. Launch SAFE Browser.
2. Load a page with links to other pages.
3. Right click "open link in new tab" or CTRL + click on a link.

**Expected behaviour:**
New link will open in a tab next to currently selected tab.
**Actual behaviour:**
New link opens at end of current list of tabs - on right hand side of browser.
![image](https://user-images.githubusercontent.com/44327924/53083044-0f9c2000-34f6-11e9-9d65-60050b888e7a.png)
![image](https://user-images.githubusercontent.com/44327924/53083060-188cf180-34f6-11e9-81bf-6ce929f87769.png)

hunterlester:
Making this issue dependent on #593 
Tabs definitely need unique ID's for this task, otherwise views are crashing.

S-Coyle:
Verified that this is now resolved on the dev branch (tested on Ubuntu)

Issue #557, opened 2019-01-25
Review, plan, and overhaul tab behavior


hunterlester:
Related issues: #522, #520, #467, #434, #542, #457, #431, #529, #546, #539, #484, #450, 

Before addressing these issues (some of which are closed), I think we should first agree on expected tab behavior so that we know what to test for; also including tests for all the edge cases that QA has found.

Some of these edge cases are due to unexpected `webview` behavior. For example, `webview` may trigger several `did-fail-load` events with a `0` error code and `OK` error description during the loading process, although loading hasn't failed.

I'd like to more fully understand the differences between `did-stop-loading`, `did-finish-load`, and `did-fail-load` events, because although the documentation is straightforward to understand, the actual behavior is not and that is a cause for some of our unexpected tab behavior.

joshuef:
@hunterlester I think part of tabbing issues arises from using `activeTab<whatever>` actions. I think we need to squash this and _always_ pass a tab index. (Which in turn means that each tab needs _its own_ address bar etc, (right now there's one that's updated).

I'd do that probably first.

So checking for things like, `going back triggers back on the correct tab`, and cycling tabs and updating the address triggers correct tab window.

We'll also need to ensure that `updateTab`, if it changes a tabs active status, that that updates whatever tab was considered active before... (which is handled by `setAsActiveTab` right now, but not by `updateTab`...

....

Hmmm, there's some other considerations that don't come to me now. But I'll come back and note down any that do 👍

joshuef:
- [ ] History (separate to tabs current history)
- [ ] http/s behaviour (do we leave tab open... and provide a 'goback' button)
- [ ] closing tab behaviour
- [ ] how would tabIndex behaviour work when you can move tabs around?

hunterlester:
Once #571 is merged, I'll be reviewing this task again to look at what needs to be completed next.
@joshuef I've forgotten the issue that we're experiencing with tab history. Why is it that we need we a complete history of tab navigation in addition to tabs own current history?

calumcraig:
If I have two or more SAFE browser windows open and I hit a SAFE URL in a clearnet browser, the link opens in all browser windows:
![image](https://user-images.githubusercontent.com/44327924/54429215-c6677680-4717-11e9-9cb6-c76d67f79501.png)
Also, if I close a tab in one of the browser windows, it closes all instances:
[...]

PR #643, opened 2019-03-25
Create Unique TabIds and _always_ require them for actions [WIP]

### LINKED ISSUE #593: Create Unique TabIds and _always_ require them for actions
## Current Behavior

We still assume an intended tab for some actions via `activeTab` (some efforts were made to improve this in #567).

Ideally we should always be passing a tab identifier to actions, so that it's _definitively_ occurring with the tab we want.

## Possible Solution

- [ ] Generate a unique tabId for each tab on creation.
- [ ] _always_ use this for identifying tabs in actions.
- [ ] Remove _getActiveTab_ from actions that are not setting a _new_ activeTab.

manavbp:
Resolves #593

### DIFF
--- a/__tests__/reducers/windows.spec.ts
+++ b/__tests__/reducers/windows.spec.ts
@@ -1,0 +1,1121 @@
+/* eslint-disable */
+import windows from '$Reducers/windows';
+import { TYPES } from '$Actions/windows_actions';
+import initialState from '$Reducers/initialAppState';
+import { isRunningUnpacked } from '$Constants';
+
+describe( 'windows reducer', () =>
+{
+    const tabId = Math.random().toString( 36 );
+    const tabId1= Math.random().toString( 36 );
+    const tabId2 = Math.random().toString( 36 );
+    const tabId3 = Math.random().toString( 36 );
+    const IntialState = initialState.windows;
+    const basicwindow =
+    {
+        activeTab: null,
+        ui: 
+        {
+            settingsMenuIsVisible: false
+        },
+        tabs: [],
+    };
+    it( 'should return the initial state', () => {
+        expect( windows( undefined, {} ) ).toEqual( initialState.windows );
+    } );
+    describe( 'ADD_WINDOW', () => {
+        it( 'should handle adding a window',()=>{
+            const addWindow = windows( IntialState ,{
+                type: TYPES.ADD_WINDOW,
+                payload : { windowId: 1 }
+            } );
+            expect(
+                addWindow
+            ).toEqual( {
+                openWindows: 
+                {
+                    '1' : {
+                        ...basicwindow
+                    }
+                },
+                closedWindows : 
+                {
+                    '1': 
+                    {
+                        closedtabs : [],
+                        lastActiveTabs : []
+                    }
+                }
+            } );
+        } );
+        it( 'should handle adding another window',()=>{
+            const state =
+            {
+                openWindows: 
+                {
+                    '1' : {
+                        ...basicwindow
+                    }
+                },
+                closedWindows : 
+                {
+                    '1': 
+                    {
+                        closedtabs : [],
+                        lastActiveTabs : []
+                    }
+                }
+            }
+            const addWindow = windows( state,{
+                type: TYPES.ADD_WINDOW,
+                payload : { windowId: 2 }
+            } );
+            expect(
+                addWindow
+            ).toEqual( {
+                openWindows: 
+                {
+                    '1' : 
+                    {
+                        ...basicwindow
+                    },
+                    '2' : {
+                        ...basicwindow
+                    }
+                },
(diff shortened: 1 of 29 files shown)


app/actions/winodw_actions.ts
```diff
@@ -0,0 +1,20 @@
...
+export const {
+    WindowaddTab
+    , WindowsetActiveTab
+    , WindowcloseTab
+    , WindowreopenTab
+} = createActions(
+    TYPES.WINDOW_ADD_TAB
+    , TYPES.WINDOW_SET_ACTIVE_TAB
+    , TYPES.WINDOW_CLOSE_TAB
+    , TYPES.WINDOW_REOPEN_TAB
```
joshuef: window / open/close menu?

app/reducers/tabs.ts
```diff
@@ -1,263 +1,94 @@
-
-import { remote, shell, webContents } from 'electron';
+/* eslint-disable */
```
joshuef: this disabled it for the entire file, I think :(

app/reducers/windows.ts
```diff
@@ -0,0 +1,212 @@
...
+/**
+ * Get the current window's webcontents Id. Defaults to `1` if none found.
+ * @return { Integer } WebContents Id of the curremt BrowserWindow webcontents.
+ */
+const getCurrentWindowId = () =>
+{
+    let currentWindowId = 1; // for testing
+    if ( remote )
+    {
+        currentWindowId = remote.getCurrentWindow().webContents.id;
```
joshuef: consider how this will work with tests (where there is not necessarily a window)

app/reducers/windows.ts
```diff
@@ -0,0 +1,212 @@
...
+        currentWindowId = currentWindow.id;
+    }
+    return currentWindowId;
+};
+
+
+const windowAddTab = (state, tab) => 
+{
+    const openWindows = { ...state.openWindows };
+    const targetwindowId = tab.windowId || getCurrentWindowId();
```
joshuef: Maybe we should not have this fallback of `getCurrentWindowId`, and _enforce_ passing of a window ID? That might prevent bugs

app/reducers/windows.ts
```diff
@@ -0,0 +1,212 @@
...
+    if (openWindows[targetwindowId]) {
+      openWindows[targetwindowId].tabs.push(tabId);
+      newWindow = {
+        ...openWindows[targetwindowId],
+        activeTab: tabId,
+        tabs: openWindows[targetwindowId].tabs
+      };
+    } 
+    else if(openWindows[targetwindowId] === undefined){
+      newWindow = {
```
joshuef: I'm not sure if we should be adding open windows here, or throwing an error?

manavbp:
> I guess this is just one of many commits for this PR too, right @manavbp ? The actually updating of code to do this will follow?
Yep, So that the PR can be broken down into a few smaller commits!

__tests__/reducers/windows.spec.ts
```diff
@@ -44,16 +45,74 @@ describe( 'windows reducer', () =>
...
+                closedWindows : 
+                {
+                    '1': 
+                    {
+                        closedtabs : [],
+                        lastActiveTabs : []
+                    }
+                }
+            }
+            const addTabNext = windows(JSON.parse(JSON.stringify(state)),{
```
joshuef: You're creating a new state just above. I don't think the parse/stringify combo is needed here...?

app/reducers/windows.ts
```diff
@@ -0,0 +1,224 @@
...
+  return newState;
+};
+
+const closetab = (state,tab) =>
+{
+  const targetWindow = tab.windowId;
+  const openWindows = { ...state.openWindows };
+  const closedWindows = { ...state.closedWindows };
+  const tabId = tab.tabId;
+  const lasttabIndex = openWindows[targetWindow].tabs.findIndex(tab => {return tab === tabId});
```
joshuef: `lastTabIndex`

joshuef:
@manav wasn't https://github.com/maidsafe/safe_browser/issues/599 to be sorted by this issue?

- [x] this does fix: https://github.com/maidsafe/safe_browser/issues/636

I would also say this issue can be the final nail in the coffin over here too: https://github.com/maidsafe/safe_browser/issues/557

joshuef:
@manavbp select nextTab/previousTab (from menu, or via shortcut) is not working

joshuef:
Also having many tabs open and switching between them causes issues where the active tab is no longer rendered

joshuef:
I can sort the tab switch/rendering issue by switching active tabs to use `z-index` away from `visibility`. 

(Though I feel like that's going backwards..., as that was an olde issue)

joshuef:
closing as this is superseded by #751

PR #678, opened 2019-04-16
fix(auth-app): adds necessary keys to resolve error message

### LINKED ISSUE #498: Some warnings are seen in the browser dev tools console
After logging in the authenticator, some 
react warnings are shown in the dev tools console which shall be removed

screenshot from 2018-12-26 17-55-51

hunterlester:
Resolves #498 

QA:
- Upon logging in to authenticator, there should no longer be error from from React about missing `key` props.


safesurfer:
I don't see any missing `key` props error from React when I log in to the authenticator. But while opening Browser I see the following error similar to this on Windows. Is this OK? @hunterlester 

image

Issue #591, opened 2019-02-15
Lint FFI files based upon updated lint rules


manavbp:
FFI files of the browser have to be updated based upon updated lint rules

joshuef:
https://github.com/maidsafe/safe_browser/pull/737 effectively closes this for now.

safesurfer:
QA verified and moving it to Done

PR #680, opened 2019-04-17
chore(lint): Resolves no-unused-vars

hunterlester:
- Turns off prevent-abbreviations lint rule
  - Turns off no-default-export lint rule

Tracking #591


joshuef:
I'm pretty in favour of `no-default-export`, so let's have a chat about this in the next meet.

@hunterlester if you want to remove that change then I'm good to get the prevent-abbr in!

Issue #570, opened 2019-02-01
Authenticating and xdg-open


b-zee:
Due to a bug in `xdg-open` a registered protocol scheme like `safe-alphachars01234` won't be opened correctly due to the digits in there. https://bugs.freedesktop.org/show_bug.cgi?id=104973

Also, because the desktop entry files containing quoted values in the `Exec` key --- like `Exec="/usr/bin/safe-browser" %u` --- `xdg-open` won't be able to open the browser binary. (https://askubuntu.com/q/189822)

Also see this topic on SAFE Dev Forum about users encountering the issue: https://forum.safedev.org/t/app-webfetch-is-not-a-function-with-safe-browser-built-from-source/2301

joshuef:
Nice research @b-zee.

We could (potentially) negate the xdg issue with `opn` perhaps (as that ships with its own), assuming that's up to date... 

Coupled with the string escaping this might solve a few xdg type issues 👍

hunterlester:
I'm using `xdg-open` v1.1.2, which I see does not include the most recent [commit](https://cgit.freedesktop.org/xdg/xdg-utils/commit/?id=bfcefa162b1dcd6d62e193019969ff2f5ff331cf) to support digits in uri scheme regex.
Why am I not able to reproduce the issue?
I'm able to `xdg-open` on `safe-bmv0lm1hawrzywzllnnhzmutynjvd3nlcg`, which does include digits.

hunterlester:
@b-zee 
> Also, because the desktop entry files containing quoted values in the Exec key --- like Exec="/usr/bin/safe-browser" %u --- xdg-open won't be able to open the browser binary.

Under what conditions though are you seeing an error?

Is `Exec="browser/node_modules/electron/dist/electron" "/home/hunterlester/safe/browser/app/main.prod.js" %u` giving you issues?

b-zee:
> Under what conditions though are you seeing an error?

The conditions are the same as mentioned in the AskUbuntu website.

b-zee:
As the issue is really with `xdg-utils` I wanted to mention that a fix has been committed a while ago (March 2019) by XDG: https://gitlab.freedesktop.org/xdg/xdg-utils/commit/bfcefa162b1dcd6d62e193019969ff2f5ff331cf

It's now about waiting for them to release a new version of their packages containing the fix.

joshuef:
with auth moving out of the browser I'm closing this

PR #681, opened 2019-04-18
chore(deps): use open instead of shell.openExternal

hunterlester:
Tracking #570 

`open` is actively maintained, so although this `xdg-utils` [commit](https://cgit.freedesktop.org/xdg/xdg-utils/commit/?id=bfcefa162b1dcd6d62e193019969ff2f5ff331cf) is not yet stable released, it will be and `open` maintainers will be on top of it.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -85,7 +85,7 @@ before_script:
 
 script:
 # update the xdg implementation on the machine. older version is buggy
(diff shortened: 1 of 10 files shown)


joshuef:
working on osx. still need to check linux

joshuef:
seems good on linux

PR #684, opened 2019-04-22
fix(ConnectedRouter): Previously connected-react-router was listed a …

joshuef:
Moves connected-react-router to be package dependency (not dev dep).

QA: Normal regression tests only.


joshuef:
Following up: https://safenetforum.org/t/bug-during-developing-app/28376/11 I noted that the error was being thrown due to the dep being listed as a `dev dep` only.

This fixes that.

PR #694, opened 2019-04-25
fix(Invite): Invite page not notified as blocked when opening in clea…

joshuef:
QA: 

Click on the invite link. The page should open in the clearnet browser, and there should be _no_ warning about the blocked url.

Sadly the tab still hangs around, but that's the standard behaviour we have here atm.

### DIFF
--- a/app/extensions/safe/blockNonSafeReqs.ts
+++ b/app/extensions/safe/blockNonSafeReqs.ts
@@ -1,5 +1,5 @@
 import open from 'open';
-import { remote, shell } from 'electron';
+import { remote } from 'electron';
 import { parse as parseURL } from 'url';
 import path from 'path';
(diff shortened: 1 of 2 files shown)


app/extensions/safe/blockNonSafeReqs.ts
```diff
@@ -17,43 +17,49 @@ const blockNonSAFERequests = () => {
...
-            if ( allowedHttp.includes( details.url ) ) {
-                try {
-                    open( details.url );
-                } catch ( e ) {
-                    logger.error( e );
+            if ( httpRegExp.test( details.url ) ) {
+                if ( allowedHttp.includes( details.url ) ) {
+                    try {
+                        open( details.url );
+                        callback( { redirectURL: 'about:blank' } );
```
hunterlester: Just needs `return` after this line so that `callback` is not called more than once.

PR #713, opened 2019-04-30
feat(context): Add more context menus.

### LINKED ISSUE #596: F5 for reload page
**Feature request**

Currently the SAFE browser uses either the refresh button or CTRL + R as options to reload the pages - these are common (Firefox and Chrome anyway) browser navigation methods.

Can we add simply F5 to reload - again both Chrome and Firefox currently have this option. (From a personal point of view this is the method I prefer to use to reload a page.)

### LINKED ISSUE #597: Right click for navigation options (back / forward / reload)
**Feature request**

In Firefox and Chrome, right mouse click gives a context menu which includes the navigation options of back, forward and reload:
Chrome:
image
Firefox:
image
Currently the only right click option in SAFE Browser is "Inspect element" - can navigation options (which exist as buttons / menu items) be added to the right click menu?

joshuef:
Fixes #597


joshuef:
I'd referenced the wrong issue there. Commit message update, and the real target is #597

PR #723, opened 2019-05-02
Fix opening urls

### LINKED ISSUE #720: External SAFE links not opening consistently after first instance of SAFE browser has been closed.
**Steps to reproduce:**
1. Launch SAFE browser and open 1+ browser windows.
2. Close the first instance of the browser (the window with the authenticator).
3. Click on SAFE links in regular browser.

**Expected behaviour:**
Links will open the corresponding SAFE sites in the SAFE browser when clicked.
**Actual behaviour:**
Links do not open immediately. If user _only_ clicks on SAFE links in regular web browser the SAFE sites will not open. User needs to click _back_ to the SAFE browser before SAFE links open.

joshuef:
Fixes #720

### DIFF
--- a/__tests__/reducers/windows.spec.ts
+++ b/__tests__/reducers/windows.spec.ts
@@ -12,8 +12,9 @@ describe( 'windows reducer', () => {
     const firstWindowId = '1';
     const secondWindowId = '2';
     const IntialState = initialAppState.windows;
-    const basicwindow = {
+    const basicWindow = {
         activeTab: null,
+        wasLastInFocus: true,
         ui: {
             settingsMenuIsVisible: false
         },
@@ -32,7 +33,7 @@ describe( 'windows reducer', () => {
             expect( addWindow ).toEqual( {
                 openWindows: {
                     [firstWindowId]: {
-                        ...basicwindow
+                        ...basicWindow
                     }
                 },
                 closedWindows: {
@@ -47,7 +48,7 @@ describe( 'windows reducer', () => {
             const state = {
                 openWindows: {
                     [firstWindowId]: {
-                        ...basicwindow
+                        ...basicWindow
                     }
                 },
                 closedWindows: {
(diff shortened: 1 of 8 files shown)


hunterlester:
@joshuef Working as expected on Linux.
I'm spending time to understand the issue on Windows.

hunterlester:
What's odd is that I see the same error as is output on `dev` branch:
```
[main 20:57:07.352] ΓÇ║ >>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>
[main 20:57:07.354] ΓÇ║ Unhandled Rejection. Reason: Cannot read property 'webContents' of null
[main 20:57:07.355] ΓÇ║ {"message":"Cannot read property 'webContents' of null","name":"TypeError"}
[main 20:57:07.355] ΓÇ║ undefined
[main 20:57:07.356] ΓÇ║ undefined
[main 20:57:07.356] ΓÇ║ >>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>
```

joshuef:
Ah, interesting @hunterlester , so it's still not finding the right browserwindow. 

Could throw some logs in there and see what you see for `allWindows` https://github.com/maidsafe/safe_browser/pull/723/files#diff-f4f911700ed16d64f7bb1f8e80001e85R108

joshuef:
This may well be simpler with @manavbp's upcoming tab improvement branch, where we can more simply query the store for windows.

hunterlester:
https://github.com/electron/electron/issues/984 describes the issue that we are having on Windows.

Trying `setImmediate( () => extension.onReceiveUrl( store, url ) );` here, 
https://github.com/joshuef/safe_browser/blob/Fix-OpeningUrls/app/extensions/index.ts#L164, I once saw a successful reception of system URI handling, then back to the above error.

PR #784, opened 2019-05-16
fix(resetStore): fix resetStore Action

manavbp:


joshuef:
@manavbp I noticed a couple things here:

- [ ] `disapatch` is causing an error
- [ ] `ancientHistory` is not interable (so maybe we just need `|| []` when defining...
- [ ] There's a `Tab.tsx` error re: `executeJavaScript`

Beyond that seems like appropriate things are firing to me.

PR #794, opened 2019-05-21
Feat test cafe provider

joshuef:
- Enable safe page testing in testcafe
- Adds basic authenticator page tests
- adds a check in packaged macOs to move app to applications folder if not located there.

QA: Tests passing. + Package app on MacOS and open (not in applications), you should be prompted to move app before it will load.


hunterlester:
Linux
8/18 tests failing on `yarn cafe-e2e2`.
All due to:
```
   1) Cannot obtain information about the node because the specified selector does not match any node in the DOM tree.

       > | Selector([function])
         |   .find('input')
```

PR #795, opened 2019-05-21
feat(reload): F5 for reload

### LINKED ISSUE #596: F5 for reload page

manavbp:
resolves #596

QA: 
To test this PR 
- Open Safe browser
- For Windows keypress, F5 should reload the page
- For Mac OS keypress Ctrl or Command + R should reload the page

This Issue in electron prevents us from adding more than one keyboard shortcut for a particular menu item.
https://github.com/electron/electron/issues/4758

The issue suggests a workaround for this by handling shortcuts with DOM events
But making this does not work as within the webview the keyboard shortcut does nothing. 
https://github.com/electron/electron/blob/master/docs/tutorial/keyboard-shortcuts.md

### DIFF
--- a/app/menu.ts
+++ b/app/menu.ts
@@ -329,7 +329,8 @@ export class MenuBuilder {
                 { type: 'separator' },
                 {
                     label: 'Reload',
-                    accelerator: 'CommandOrControl+R',
+                    accelerator:
+            process.platform === 'darwin' ? 'CommandOrControl+R' : 'F5',
                     click: ( item, win ) => {
                         if ( win ) {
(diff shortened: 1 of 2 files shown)


calumcraig:
Will the shortcut on Linux be F5 or CTRL + R?

PR #796, opened 2019-05-22
Seperate history

manavbp:
Seperate History from Tab History


app/components/Browser/Browser.tsx
```diff
@@ -16,6 +16,7 @@ interface BrowserProps {
     tabs: object;
     windows: object;
     windowId: any;
+    history?: Array<any>;
```
joshuef: is this optional? It's coming from the store so should at least be an empty array or?

app/reducers/history.ts
```diff
@@ -0,0 +1,40 @@
+import { initialAppState } from './initialAppState';
+import { makeValidAddressBarUrl } from '$Utils/urlHelpers';
+import { TYPES as TABS_TYPES } from '$Actions/tabs_actions';
+import { logger } from '$Logger';
+
+const initialState = initialAppState.history;
+
+const updateHistory = ( state, payload ) => {
+    if ( payload.url ) {
```
joshuef: if no url? should we throw an error on calling that action? To alert whoever is calling the action they've done it wrong?

app/reducers/history.ts
```diff
@@ -0,0 +1,40 @@
...
+        const url = makeValidAddressBarUrl( payload.url );
+        if ( url !== state[state.length - 1] ) {
+            const newState = [...state];
+            newState.push( url );
```
joshuef: Do we only update history if it's not the same as the last one?

What if it's the same as one 5 sites before? Does that _previous_ site remain? Do we still show it if it does? (Or do we only show the latest instance? (I realise this isn't all reducer specific)

app/reducers/history.ts
```diff
@@ -0,0 +1,40 @@
...
+    if ( action.error ) {
+        logger.error( 'ERROR IN ACTION', action.error );
+        return state;
+    }
+
+    switch ( action.type ) {
+        case TABS_TYPES.UPDATE_TAB: {
+            return updateHistory( state, payload );
+        }
+        case TABS_TYPES.TABS_RESET_STORE: {
```
joshuef: Should we use a tabs action here?

Should there be a general action for resetting the store if it can be applied across many reducers? (So maybe there's 'trigger_reset' which is the alias and 'reset_store' which can be passed into any reducer?

joshuef:
@manavbp right now it looks like:

- [ ] history is not sorted newest to oldest when displayed
- [ ] back/forwards dont affect history when they should

hunterlester:
@manavbp I'm not sure why your tests are failing locally but that's not the issue in Travis because all tests are passing there. (There is a bit of linting to clean up though).
Failing due to: https://travis-ci.com/maidsafe/safe_browser/jobs/203494199#L2696

app/components/PerusePages/History/History.tsx
```diff
@@ -28,39 +28,73 @@ export class History extends Component<HistoryProps, {}> {
...
+                if (
+                    ignoreList.includes( url ) ||
+          ignoreProtocolList.includes( urlObj.protocol )
+                ) {
+                    return false;
+                }
+                return urlIsValid( url );
+            } );
+
+            if ( list.length > 1 ) {
```
hunterlester: `list.length >= 1`

__tests__/components/History.spec.tsx
```diff
@@ -34,73 +36,76 @@ describe( 'History Component', () => {
...
-
-        it( 'should have one url list', () => {
-            expect( wrapper.find( UrlList ).length ).toBe( 1 );
-        } );
-
-        it( 'should have one link', () => {
             wrapper = mount( <History {...props} /> );
-            expect( wrapper.find( 'a' ).length ).toBe( 1 );
+            it( 'should have one link', () => {
+            expect( wrapper.find( 'a' ).length ).toBe( 0 );
```
hunterlester: Revert back to `1`

__tests__/components/History.spec.tsx
```diff
@@ -34,73 +36,76 @@ describe( 'History Component', () => {
...
-        it( 'should have one url list', () => {
-            expect( wrapper.find( UrlList ).length ).toBe( 1 );
-        } );
-
+    /* 
+        ** Commented out because its failing **
         it( 'should have one link', () => {
             wrapper = mount( <History {...props} /> );
-            expect( wrapper.find( 'a' ).length ).toBe( 1 );
+            expect( wrapper.find( 'a' ).length ).toBe( 0 );
```
hunterlester: Revert back to `1`

__tests__/components/History.spec.tsx
```diff
@@ -34,73 +36,76 @@ describe( 'History Component', () => {
...
 
         it( 'should have one link with text', () => {
-            wrapper = mount( <History {...props} /> );
             expect( wrapper.find( 'a' ).text() ).toBe( 'safe://somethingreal' );
         } );
     } );
+    */
 
     describe( 'props', () => {
         describe( 'tabs', () => {
```
hunterlester: I'd opt to remove this block altogether since this component does not receive a `tabs` prop.

__tests__/reducers/history.spec.ts
```diff
@@ -0,0 +1,165 @@
...
+                url: 'helloAgain'
+            }
+        } );
+
+        expect( newState ).not.toStrictEqual( state );
+        expect( newState ).toEqual( {
+            [date]: [
+                {
+                    url: 'safe://helloAgain',
+                    timeStamp: new Date().toLocaleTimeString()
```
joshuef: @manavbp are we store localeTimeStrings here? I think we should be using UTC in the store, and converting for display only. Otherwise we should be keeping the timezone info etc too.

UTC in the store seems simplest to me. And we can let the UI figure out displaying this info, and so we don't lose any data in the store itself. (We should accordingly check that a UTC or `Date` object is passed in the store and throw otherwise (and have tests for this too)

app/components/PerusePages/History/History.tsx
```diff
@@ -28,39 +28,73 @@ export class History extends Component<HistoryProps, {}> {
 
     render() {
         const { history, isActiveTab, windowId, addTabEnd } = this.props;
-        let historyList = [];
-        const historyArray = [];
-        // eslint-disable-next-line array-callback-return
```
joshuef: why are we disabling this and not using what's recommended here?

app/components/PerusePages/History/History.tsx
```diff
@@ -28,39 +28,73 @@ export class History extends Component<HistoryProps, {}> {
...
-            const urlObj = parse( url );
-            if (
-                ignoreList.includes( url ) ||
-        ignoreProtocolList.includes( urlObj.protocol )
-            ) {
-                return false;
+        const dates = Object.keys( history );
+        let list = [];
+        const parsedList = [];
+        /* eslint-disable array-callback-return */
```
joshuef: same as above

joshuef: @manavbp also when disabling please try and stick to line only disabled (maybe there's a rule to avoid disabling on a block?)

app/reducers/tabs.ts
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable @typescript-eslint/explicit-function-return-type */
```
joshuef: again, lets do this only for specific lines, and only when it makes sense. Thus far we could just be using an arrow func and returning `null` eg.

app/reducers/tabs.ts
```diff
@@ -116,19 +117,112 @@ const updateTabHistory = ( tabToMerge, payload ) => {
...
+    logger.error( 'No URL passed' );
+    return state;
+};
 
-    const updatedState = { ...state, [tabId]: updatedTab };
-    return updatedState;
+const updateTabWebId = ( state, payload ) => {
+    const { tabId, tabToMerge } = handleTabPayload( state, payload );
+    if ( tabId === undefined && tabToMerge === undefined ) {
+        logger.error( 'tab  Does not exist' );
```
joshuef: watch capitalisation / spacing here

app/reducers/history.ts
```diff
@@ -8,9 +8,11 @@ const initialState = initialAppState.history;
 const updateHistory = ( state, payload ) => {
     if ( payload.url ) {
         const historyState = { ...state };
-        const date = new Date().toLocaleDateString();
+        const d = new Date();
+        const date = d.toLocaleDateString();
         const url = makeValidAddressBarUrl( payload.url );
-        const timeStamp = new Date().toLocaleTimeString();
+        const timeStamp = `${d.getUTCHours()}:${d.getUTCMinutes()}:${d.getSeconds()}`;
```
joshuef: just `new Date().getTime()`. Anything else is formatting to be done by frontend.

app/reducers/tabs.ts
```diff
@@ -144,7 +143,7 @@ const updateTabUrl = ( state, payload ) => {
 const updateTabWebId = ( state, payload ) => {
     const { tabId, tabToMerge } = handleTabPayload( state, payload );
     if ( tabId === undefined && tabToMerge === undefined ) {
-        logger.error( 'tab  Does not exist' );
+        logger.error( 'Tab Does not exist' );
```
joshuef: `Tab does...`

PR #843, opened 2019-06-17
chore(tabBar): Update TabBar css to prevent text overflow

manavbp:
In the packaged app on loading patter, the tab look like :-

56902837-7b0fea00-6a71-11e9-9154-c164d04d6fd6

This Pr fixes this issue.

QA-
To test this PR package the app and go to safe://patter.dapp

### DIFF
--- a/app/components/TabBar/tabBar.css
+++ b/app/components/TabBar/tabBar.css
@@ -10,7 +10,8 @@
     -webkit-app-region: drag;
 }
 .tabRow {
-    flex-wrap: nowrap;
+    display: flex;
+    justify-content: flex-start;
 }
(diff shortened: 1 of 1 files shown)


manavbp:
> @manavbp we shouldn't need to use `!important` at all. This means we're doing CSS wrong.
> 
> I think the real issue here is, why doesn't `nowrap` work on the container? It should not allow this. Is it in the correct place?

@joshuef the reason we are using is because antd-col sets the position to relative thereby causing the postion of the favicon and close icon to be miss aligned. Adding !important overrides antd style class and sets postion to absolute. Which is why !important is required.

PR #844, opened 2019-06-17
chore(package): Provide a AppImage for linux

manavbp:
Packaging the App on Linux should provide an app Image.

QA-
The best way to test this app is to package the app on Linux and check if an AppImage is Provided to as an executable.

### DIFF
--- a/package.json
+++ b/package.json
@@ -149,7 +149,7 @@
       "target": "dir"
     },
     "linux": {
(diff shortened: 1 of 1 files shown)


probonopd:
Is the AppImage available somewhere for download? Please do not put it inside an archive like zip or tar.gz.

PR #856, opened 2019-06-21
Fix for History TimeStamps

### LINKED ISSUE #854: History timestamps have both AM / PM and are in 24 hour clock format.
Expected:
The time visited for pages in the history record is recorded either as (eg) 3:00 PM _or_ 15:00.
Actual:
The time visited for pages in the history record is recorded with **both** 12 and 24 hour clock styles.
image

manavbp:
resolves #854

### DIFF
--- a/app/components/PerusePages/History/History.tsx
+++ b/app/components/PerusePages/History/History.tsx
@@ -119,6 +119,9 @@ export class History extends Component<HistoryProps, {}> {
                             <H1 title="History" />
                         </PageHeader>
                     </Page>
+                    <h6 className={`${styles.warning}`}>
+            *Note: Time displayed is in GMT format
+                    </h6>
                     <Table className={styles.table}>
                         {parsedList}
                         {!parsedList.length && (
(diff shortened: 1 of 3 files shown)


app/components/PerusePages/History/History.tsx
```diff
@@ -114,6 +119,9 @@ export class History extends Component<HistoryProps, {}> {
                             <H1 title="History" />
                         </PageHeader>
                     </Page>
+                    <h6 className={`${styles.warning}`}>
+            *Note: Time displayed is in GMT format
```
joshuef: I dont think we need this warning TBH.

manavbp: We add the warning to notify users about the timeStamps being in GMT format, while they would usually expect a local timeStamp.

joshuef: So, I think GMT is a timezone and not a format. You're looking for 24hr of 12hr format. (12hr being w/ AM/PM).

But I also don't think we need a note for this.

Issue #852, opened 2019-06-20
Launching experimental apps when experimental API mode is not switched on does not give warning


calumcraig:
**Steps to reproduce:**
Testing SAFE Browser v0.14-RC1 - macOS.
1. Launch browser and log into SAFE network.
2. Do NOT enable experimental mode.
3. Launch a web app that requires experimental API. eg WebID manager or Patter.

**Expected:**
An error message advising that experimental mode must be enabled is displayed (as previously happened).
**Actual:**
App requests authorisation but no error related to experimental mode is displayed.

S-Coyle:
Closing legacy issues

PR #874, opened 2019-06-25
Pr#852

### LINKED ISSUE #813: Issue with `playground.dapp` which is not allowing me to pass `safeApp.authorise`, though it returns `authUri` in the terminal
STR:

1. Open SAFE Browser
2. Open `playground.dapp`, and `initialiseApp`, `genAuthUri`
3. Now try to `authorise`, click `allow` in the auth popup

It returns the `authUri` in the terminal, but it doesn't get passed to the next step and kept on loading.

image

manavbp:
Resolves #813 #852


manavbp:
@joshuef I think having these changes added should be a separate issue

PR #884, opened 2019-07-01
Remove duplication of functionality between webPreload and webviewPreload

### LINKED ISSUE #875: Remove duplication of functionality between webPreload and webviewPreload
The duplication of this functionality we have between webPreload and webviewPreload (also maybe rename the latter file to make it clearer it is the SAFE functionality executed on preload?)

To do-

- [x] Remove duplicate code between WebViewPreload and WebPreload
- [x] Move createRemoteCall from WebViewPreload to WebPreload

manavbp:
resolves #875 

QA-
To test this PR
1) load the safe browser 
2) Check if the dom Apis are loaded in
3) Login to the browser
4) Go to safe://playground.dapp
5) Test if you can authorize the app

### DIFF
--- a/app/extensions/safe/test/app/webviewPreload.spec.ts
+++ b/app/extensions/safe/test/app/webviewPreload.spec.ts
@@ -1,4 +1,5 @@
 import * as webviewPreload from '$Extensions/safe/webviewPreload';
+import { addRemoteCall } from '$Actions/remoteCall_actions';
 import { APP_INFO, startedRunningProduction } from '$Constants';
 
 // avoid appveyour for its weak.ref issues right now.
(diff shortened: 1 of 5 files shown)


app/extensions/safe/test/app/webviewPreload.spec.ts
```diff
@@ -16,6 +17,34 @@ jest.mock( 'extensions/safe/ffi/authenticator', () => jest.fn() );
 
 jest.mock( '@maidsafe/safe-node-app', () => jest.fn() );
 
+const pendingCalls = {};
+
+const createRemoteCall = ( functionName, passedStore ) => {
```
joshuef: Can we not pull this in from `webPreload`? Or mock it? Seems overkill / dangerous to be replicating the func here.

S-Coyle:
Closing due to PR pruning - this can be picked up again at a later date if required

PR #1016, opened 2019-10-04
Feat relative rendering of content for version

### LINKED ISSUE #1010: Browser should request relative resources from _same_ version of site
Steps to reproduce:
- create a site with an image & an index.html which will display that image on the site.
- edit that site & upload a different folder with different files or images.
- scroll back to the previous version - the original image you uploaded will not display

image

Using the cli we are able to use `cat` and see/download the image still, so it is there, however the browser is not rendering it any longer

### LINKED ISSUE #1017: using versioned url doesn't inform of new versions
Going here: safe://comeonyouboysingreen/?v=1

no newer version button is available, even though its not the latest version.

WANTED:

The UI should reflect that newer versions are available

joshuef:
fixes #1010 
fixes https://github.com/maidsafe/safe_browser/issues/1017


theWebalyst:
Want :grin:

manavbp (approved):

PR #1055, opened 2019-10-22
chore(refactor): Add SNAPP update logic to browser

manavbp:
contains all the update logic required for SNAPP auto-update process


app/autoUpdate.ts
```diff
@@ -16,6 +17,11 @@ const clearNotification = ( payload: { id: string } ) =>
 
 // Thow error incase there is an issue in updating
 autoUpdater.on( 'error', ( error ) => {
+    if ( process.platform === 'linux' || process.platform === 'win32' ) {
+        updateSNAPP( [`--update-failed`, `--appId:safe.browser`] );
```
joshuef: can we pass the error message across here too? Or where would this info be shown?

app/autoUpdate.ts
```diff
@@ -24,122 +30,134 @@ autoUpdater.on( 'error', ( error ) => {
...
-    const title = 'Update SAFE BROWSER';
-    const body = 'Download Latest Version?';
-
-    const ignoreRequest = () => {
-        logger.info( 'replace these ipcRenderer.send calls' );
-        clearNotification( { id: notificationId } );
-    };
-
-    const success = () => {
+    if ( !process.argv.includes( '--trigger-update' ) ) {
```
joshuef: lets make a const for this incase we need it elsewhere.

app/main.dev.ts
```diff
@@ -123,6 +124,11 @@ app.on( 'ready', async () => {
             if ( commandLine.length >= 2 && uri ) {
                 onReceiveUrl( store, uri );
             }
+
+            if ( commandLine.includes( '--trigger-update' ) ) {
```
joshuef: maybe even we need a global var for this, as this state could be different to process.argv etc, no?

app/main.dev.ts
```diff
@@ -123,6 +124,11 @@ app.on( 'ready', async () => {
...
+                // eslint-disable-next-line no-new
+                new AppUpdater( store );
```
joshuef: Do we need a `new` instance? Can the old one be reused?

app/main.dev.ts
```diff
@@ -144,9 +150,19 @@ app.on( 'ready', async () => {
...
 
-    await setupBackground();
+    if ( !process.argv.includes( '--trigger-update' ) ) {
+        await setupBackground();
 
-    mainWindow = openWindow( store );
+        mainWindow = openWindow( store );
+    }
+
+    if ( await checkIfAppIsRunning() ) {
```
joshuef: `checkIfAppIsRunning` seems badly named. What do we want / expect to be returned here? Which app? the browser check is done via the `getInstanceLock`...
 
Maybe `XIsRunning` is a better name? (maybe not)

app/updateSNAPP.ts
```diff
@@ -0,0 +1,135 @@
...
+
+    logger.error( 'installedPath', installedPath );
+
+    return installedPath;
+};
+
+export const updateSNAPP = ( arg ) => {
+    // We could pass the app object as an arg also if needed
+    // To enable more cross app communication
+    logger.error( 'updateSNAPP called with arg', arg );
```
joshuef: `error`?

app/updateSNAPP.ts
```diff
@@ -0,0 +1,135 @@
...
+        }
+    }
+    return applicationExecutable;
+};
+
+const getInstalledLocation = ( application ): string => {
+    const applicationExecutable = getApplicationExecutable( application );
+    const installedPath = path.join( getAppInstallDir(), applicationExecutable );
+
+    logger.error( 'installedPath', installedPath );
```
joshuef: error?

app/updateSNAPP.ts
```diff
@@ -0,0 +1,135 @@
...
+export const checkIfAppIsRunning = async () => {
+    // We could pass the app object as an arg also if needed
+    // To enable more cross app communication
+    const application = {
+        name: 'SAFE Network App',
+        packageName: 'safe-network-app'
+    };
+    const appList = await psList();
+    const appIsRunning = appList.find(
+        ( element ) => element.name === `${application.name}.exe`
```
joshuef: how would this work on mac/linux?

manavbp: The package works on all three platforms and returns a Promise<Array> with the running processes.

app/updateSNAPP.ts
manavbp: Yep, that has to be changed. I was using that to make certain logs more distinguishable

app/main.dev.ts
manavbp: No this checking if SNAPP is running and only then send a response to SNAPP. Cause otherwise it would always open SNAPP

app/main.dev.ts
manavbp: I don't think this would create a new Instance as app Update takes place on the main process only

app/autoUpdate.ts
manavbp: In the browser we throw a dialog box. Displaying the error, even though the window is still hidden. On Safe-network-apps side Im not sure if we should show the error there, but rather only say app-update-failed?

joshuef: >In the browser we throw a dialog box. Displaying the error, even though the window is still hidden.
That sounds like terrible UX. An error from something I don't know is running...

We should for sure show some error indicative of the failure. But will we want to know the actual reason? Maybe not in the UI, but maybe for logging eg? If it's possible to pass the error message across, perhaps that's useful

app/main.dev.ts
joshuef: So let's name it appropriately (also considering the context it's to be used in as a bool)

app/updateSNAPP.ts
joshuef: You can filter logs for a given string in devtools. Ideal is to get decent log messages (at appropriate levels), from the get go so we dont accidentally commit such things. (it's a lofty goal. I miss nonsense logs all the time myself :P )

app/updateSNAPP.ts
joshuef: specifically `${application.name}.exe` ?

manavbp: In this case it would Look for `SAFE Network App.exe` to check if a process is running and resturn an array of the current SNAPP running processes

PR #1068, opened 2019-10-31
Add ENV variable to allow package to use s3 test folder for update

manavbp:


app/constants.ts
```diff
@@ -17,6 +17,7 @@ let shouldRunMockNetwork = fs.existsSync(
 
 let hasDebugFlag = false;
 let triggerUpdate = false;
+let shouldOpenTestPackages = false;
```
joshuef: shouldOpen? or shouldBuild?

app/constants.ts
```diff
@@ -68,6 +69,11 @@ if ( allPassedArguments.includes( '--trigger-update' ) ) {
     triggerUpdate = true;
 }
 
+if ( allPassedArguments.includes( `--testPackages` ) ) {
```
joshuef: also should be settable via env var...

app/constants.ts
manavbp: You're right it should be shouldBuild

PR #1090, opened 2019-11-15
feat(releases): use release channels

joshuef:
Use release channels, so `-alpha` in package.json will generate alpha releases.

`-beta` beta.

Also updates bump command to pull tags from upstream/origin before doing the bump


S-Coyle (approved):
👍

PR #1166, opened 2020-01-27
fix(register-site): fix refactor's missing e->event

joshuef:


ravinderjangra (commented):
Not sure if it's required to change but I just noticed there is one more place where `e.timestamp` is using where there is no `e` object. maybe that can also be replaced with the same.
https://github.com/maidsafe/safe_browser/blob/8874deb7f0e179ec44feb6beef21a62c4309472c/app/components/Tab/Tab.tsx#L609

joshuef:
nice one @ravinderjangra . Got that and another and updated pr 🙇

S-Coyle (approved):
👍

Issue #1156, opened 2020-01-22
Release files tidy up


S-Coyle:
On a full version release there are 19 files published on the draft GitHub release:
1. alpha-linux.yml
2. alpha-mac.yml
3. alpha.yml
4. beta-linux.yml
5. beta-mac.yml
6. beta.yml
7. latest-linux.yml
8. latest-mac.yml
9. latest.yml
10. com.electron.safe-browser.plist
11. safe-browser-v0.15.4-beta.2-linux-x64.AppImage
12. safe-browser-v0.15.4-beta.2-linux-x64.zip
13. safe-browser-v0.15.4-beta.2-mac-x64.dmg
14. safe-browser-v0.15.4-beta.2-mac-x64.dmg.blockmap
15. safe-browser-v0.15.4-beta.2-mac-x64.pkg
16. safe-browser-v0.15.4-beta.2-mac-x64.zip
17. safe-browser-v0.15.4-beta.2-win-x64.exe
18. safe-browser-v0.15.4-beta.2-win-x64.exe.blockmap
19. safe-browser-v0.15.4-beta.2-win-x64.msi

From what I can see looking at the previous release (0.15.3) there are a couple of files produced here that are not used - numbers 10 and 19.
Any need to publish either of these?

Going by previous releases, once the 17 files are copied to S3, the GitHub release can be cut to only contain 3 release files:
1. safe-browser-v0.15.4-beta.2-linux-x64.AppImage
2. safe-browser-v0.15.4-beta.2-mac-x64.dmg
3. safe-browser-v0.15.4-beta.2-win-x64.exe
Can this be done as part of CI?

Another tidy up task on S3 would be to remove all previous Alpha & Beta files - once this new version is available they are no longer required.

One last request on this card is to ensure that all the files published to S3 are actually required. E.g. is the safe-browser-v0.15.4-beta.2-mac-x64.zip file ever used? If there are any that are now surplus, as the .msi file appears to be, let's take the opportunity to remove them.

joshuef:
`zip` files are used as part of autoupdate aye  👍

joshuef:
Above PR does _most_ of this. 

Missing things:

- Mac e2e on GHA. We still use travis for this _only_.
- Any s3 management of files (ie deleting old things). I'm unsure what sort of margin we should give. But maybe just delete anything older than a couple months maybe? I'd opt for that in another PR though @S-Coyle

S-Coyle:
@joshuef 
[...]

PR #1199, opened 2020-02-27
Gha

joshuef:
Automate sha checks
Publishing to s3
Release changelog generation
Correct output artefacts in release
store artefacts on each master build

For #1156


S-Coyle (approved):
👍

PR #1358, opened 2020-09-03
chore(deps): update to reference renamed sn_nodejs crate

S-Coyle:


joshuef (approved):

S-Coyle:
Added the `Next release` flag as merging this will trigger a new release, but it is not expected that a new release would be functional at this time so better to merge this once other work has been completed to make a new release compatible with latest vaults, cli, etc.

